PR #1, opened 2018-07-19
Add TabList

davidtheclark:
@colleenmcginnis for review, please.

screen shot 2018-07-18 at 5 57 56 pm


colleenmcginnis:
@davidtheclark the bottom border and bold text doesn't seem to be working on the docs site... Is that something to be concerned about? Everything else looks 👌 to me. 

image

davidtheclark:
Turned out it was a bug in the component, @colleenmcginnis. Should be fixed now. I also deleted the links example because the way links send you all over the docs page is kind of confusing; I think the one example should be sufficient.

davidtheclark:
Published in 0.2.0, @colleenmcginnis

PR #3, opened 2018-07-20
Fix build & release details

davidtheclark:
These changes should prevent me from making the same mistakes I made when I released v0.2.0.

@jseppi for review, please.

cc @mapbox/frontend-platform

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,15 @@
 {
+  "private": true,
   "name": "@mapbox/mr-ui",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "UI components for Mapbox projects",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,14 +1,15 @@
 {
+  "private": true,
```
jseppi: I am a little confused with npm's documentation on the intersection of scoped packages and this `"private"` config. Just double checking that we do want it set to `true` here.

jseppi: Oh, I see. `build-package.js` creates the publishable `package.json`, and this `"private: true"` is to prevent accidentally publishing the "development" `package.json`. 👍

PR #8, opened 2018-07-27
Improve docs styling

davidtheclark:
There's a lot of cleaning up that needs to be done in the documentation. I thought I'd start by trying to establish a nicer style.

Here are some highlights from my flailing attempt, along with screenshots with garish pink pointers:

- Link to the Mr. UI repo.
- Padding around the rendered examples (no need to show exactly how they lay out, as in the test cases).
- A little example-code-toggling button that looks kind of tabby.
- Props are in a table.
- Purple "Required" badge.
- "Default value" is beneath the description because it doesn't really need its own column, I thought.
- Prism is highlighting JSX nicely.
- Light gray lines between sections!

2018-07-26 at 5 35 pm

2018-07-26 at 5 36 pm

@danswick for review, please. After this, we can do a round of content & example cleaning before publishing the site.

@dasulit ... I hope you're doing well ...

### DIFF
--- a/src/docs/components/component-example.js
+++ b/src/docs/components/component-example.js
@@ -11,29 +11,59 @@ export default class ComponentExample extends React.Component {
   };
 
   renderCode() {
-    if (!this.state.showCode) return null;
+    if (!this.state.showCode) {
+      return null;
+    }
     return (
-      <pre className="mt6 language-js">
+      <pre className="my0 round-tl pre language-jsx">
         <code dangerouslySetInnerHTML={{ __html: this.props.code }} />
       </pre>
     );
   }
 
   render() {
-    const { props } = this;
+    const { props, state } = this;
     return (
       <div>
-        <div className="mb6 prose">{props.description}</div>
-        <div className="border round border--gray-light border--dash">
-          {React.createElement(props.exampleModule.default)}
+        <div className="flex-parent flex-parent--end-cross">
+          <div className="flex-child flex-child--grow pb6">
+            {props.description}
+          </div>
+          <div className="flex-child flex-child--no-shrink w120">
+            <div className="flex-parent flex-parent--end-main">
+              <div className="flex-child">
+                <ToggleCodeButton
+                  onClick={this.toggleCode}
+                  codeIsVisible={state.showCode}
+                />
+              </div>
+            </div>
+          </div>
         </div>
-        <div className="mt6">
-          <button className="btn btn--s" onClick={this.toggleCode}>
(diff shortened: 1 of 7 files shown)


dasulit:
Love that prop table!

src/docs/components/component-example.js
```diff
@@ -11,29 +11,59 @@ export default class ComponentExample extends React.Component {
...
-        <div className="border round border--gray-light border--dash">
-          {React.createElement(props.exampleModule.default)}
+        <div className="flex-parent flex-parent--end-cross">
+          <div className="flex-child flex-child--grow pb6">
+            {props.description}
+          </div>
+          <div className="flex-child flex-child--no-shrink w120">
+            <div className="flex-parent flex-parent--end-main">
+              <div className="flex-child">
+                <ToggleCodeButton
```
danswick: I think a `ControlToggleSet` might be good to use here to more clearly indicate the two possible states each section can be in (code vs rendered example).

src/docs/components/sidebar.js
```diff
@@ -16,8 +16,18 @@ export default class Sidebar extends React.Component {
...
-      <div className="pt24 txt-s">
-        <ul>{componentLinks}</ul>
+      <div>
+        <a
+          className="link"
+          href="https://github.com/mapbox/mr-ui"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          <h1 className="txt-bold txt-h3 txt-mono">Mr UI</h1>
```
danswick: I might use `<Icon name="github" />` or something here to indicate that the link is to an external page. Otherwise it seems to me like this link would go to a main Mr UI docs page.

src/docs/components/component-example.js
davidtheclark: I'm not sure that's a common element for expand/collapse situations like this. I'm trying to think of what words to put in the toggles: "See code / hide code" seems verbose ... Did you have some words in mind that would work?

danswick: Sounds good. I tested a rough idea yesterday and ended up with huge toggles for this reason. 

In the future, I think `example` and `code` might be self-explanatory enough.

davidtheclark: > In the future, I think example and code might be self-explanatory enough.

Ah, but we're not toggling between the two: the example is *always* visible, it's just the code that is hiding & showing.

danswick: Oh wow, not sure how I missed that 🤦‍♂️

PR #20, opened 2018-07-31
MbxButton

### LINKED ISSUE #11: Buttons!
It's time to create some opinionated button components. 

Here's what I'm thinking we'll want ...

For all buttons, there will be props that make it easy to 

- add a vertically centered icon before the text
- control colors
- stretch the button to the full width of its container
- make it a larger size — e.g. on marketing pages we provide some extra horizontal padding
- disable the button (in which it'll use Assembly's disabled style)
- make the button an icon-only button with a tooltip text explanation (replacing the need for our current IconButton component)

**MbxButton** looks like this:

screen shot 2018-07-27 at 11 13 28 am

That's what we have throughout Studio and marketing pages. Account Dashboard is the outlier with larger text. There seems to be to be enough enthusiasm for the smaller-text norm that we should stick with that.

**MbxSecondaryButton** is the outlined version of the primary button:

screen shot 2018-07-27 at 12 09 16 pm

**MbxDiscouragingButton** is gray, but not too gray, and outlined:

screen shot 2018-07-27 at 12 07 05 pm

**MbxTextButton** looks like a bold link. Usually it includes an icon, and it can take on whatever text size its container is using.

screen shot 2018-07-27 at 12 12 11 pm

**MbxAppButton** is for denser apps like Studio's style editor. Borders are not fully rounded, text is usually smaller — though we could have two sizes, `txt-xs` and `txt-s`. The default color is gray instead of blue, so it's not hogging too much attention in a dense UI. (You'll be able to make it blue with an option).

screen shot 2018-07-27 at 11 16 53 am
screen shot 2018-07-27 at 11 16 56 am

**MbxAppSecondaryButton** is the same as MbxAppButton but outlined instead of filled.

davidtheclark:
Closes #11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## HEAD
 
-- Add *MbxIconText** component.
-- Add *MbxUnderlineTabs** component.
+- Add **MbxButton** component.
+- Add **MbxIconText** component.
+- Add **MbxUnderlineTabs** component.
 - **PopoverTrigger**
   - If the trigger responds to focus but not click, and you focus the trigger *first* and *then* click, that click *closes* the popover instead of leaving it open even after you move the mouse away.
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -732,9 +732,9 @@
       "integrity": "sha512-Z5ENbYjbN5NgfITHq/o9Yspm/tx4H+LdcfnPGksb0KjK2DmRlSY4fN8vzrNLLHcMX05puARr2EU+LkFMK+UNtQ=="
     },
     "@mapbox/react-test-kitchen": {
-      "version": "0.1.2",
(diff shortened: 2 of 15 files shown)


davidtheclark:
Here's what I ended up with:

`MbxButton` is a single component that takes a `variant` prop. `variant` is a keyword that indicates a certain style of button. You can also customize the appearance through `size`, `width`, `outline`, `corner`, and `transformClasses` props. The goal was to make it easy to create the styles that I us reusing.

Here are the docs and test cases:

screen shot 2018-08-02 at 11 55 10 am
screen shot 2018-08-02 at 11 55 19 am
screen shot 2018-08-02 at 11 55 25 am
screen shot 2018-08-02 at 11 55 35 am
screen shot 2018-08-02 at 11 55 43 am
192 168 1 107_9966_mbxbutton

Issue #26, opened 2018-08-02
Tiered styling properties


davidtheclark:
Having experimented with a few different options, here's the outline of a styling pattern that I think could fit our purposes:

- Opinionated defaults.
- Standard variants are easy to implement, with either boolean or keyword props. [The current MbxButton component illustrates this approach](https://github.com/mapbox/mr-ui/blob/5b19a30721fb473c5db9e5dc9c1be623672d697f/src/components/mbx-button/mbx-button.js) with the `variant`, `size`, and `width` props.
- Non-standard variants are *possible* to implement, with `theme*` props. This is the pattern established by the bulk of existing components and has proven useful. The main problem with it is that it requires a lot of the user — so it makes sense for advanced users, but not for the average user. We should continue to use these but indicate in the docs that they are advanced props and most people should not need to use them.

One major annoyance with the `theme*` props is that you need to _replace_ every class that you _don't_ want to override. A common use case is to just *append* a class to the end of the list, but to do this you actually need to copy-paste the default class list and add your own to the end — and that's not very robust or cheery. So I'm thinking that instead of `theme*` props accepting only *strings*, they'll also accept *functions* that receive the default class list and return a modified one. Appending would be very easy, then.

cc @danswick

davidtheclark:
@kepta helped come up with an adjustment to this that will probably be nice. 

Since the most common use case for `theme*` props is probably to *append* a class or two, and only the Very Advanced User would want to replace the whole class list or remove parts of it, we can make that primary use case easier: If you provide a string to a `theme*` prop, it will be appended to the class list. If you provide a function, you can do whatever you want, wholly replacing the class list with whatever you return.

davidtheclark:
I audited our use of `theme*` props in major codebases, and I think now that we *should not* keep them around. Here's what I think transition off of them would look like:

- Common style variations will be built into the component.
- Layout can be determined with a wrapper component. Lots of `<Icon themeIcon="..." />` usage, for example, is just to achieve layout that could be done with a wrapper.
- If *complete* customization is really necessary and making a custom component isn't in the cards, because the functionality is complex, we'll try to expose *unstyled low-level components*. Essentially following the strategy of [react-aria-modal](https://github.com/davidtheclark/react-aria-modal), where you have to bring all your own styles. For example, we could expose both a styled and an unstyled tooltip or popover. That would come with the caveat that the unstyled one would rely on your *completely* for its styling.
[...]

PR #29, opened 2018-08-04
Proof-of-concept for theme props approach

davidtheclark:
Relates to https://github.com/mapbox/mr-ui/issues/26

The first commit on this PR does a few things:

- Modify the Button component to have a `theme` prop that fits the convention proposed in #26: string values are appended, function values replace.
- Create an `applyTheme` function that we can use to apply the same convention all over the place.
- Modify the documentation to de-emphasize `theme*` props, since they are for advanced users, and explain them all in one place.

If we like this approach, I guess the next step would be to modify all the existing components to follow the new convention. To be clear, it would a breaking change: usage of `theme*` strings right now would have to adjust to accommodate the change. The intended benefit of the change is to help the most common use case and the less advanced users. If you disagree and think we should just stick with the current `theme*` behavior — where `theme*` props take a string and that string replaces the class list wholesale — please let me know! cc @tristen @mapbox/studio @frontend

Here are some shots of what the docs look like:

screen shot 2018-08-03 at 5 15 21 pm
screen shot 2018-08-03 at 5 15 28 pm
screen shot 2018-08-03 at 5 15 39 pm
screen shot 2018-08-03 at 5 15 46 pm

### DIFF
--- a/src/components/button/__tests__/__snapshots__/button.test.js.snap
+++ b/src/components/button/__tests__/__snapshots__/button.test.js.snap
@@ -50,7 +50,7 @@ exports[`Discouraging renders as expected 1`] = `
 </button>
 `;
 
-exports[`Div styled like a medium destructive button with some extended props and transformed classes renders as expected 1`] = `
+exports[`Div styled like a medium destructive button with some extended props and custom classes renders as expected 1`] = `
 <div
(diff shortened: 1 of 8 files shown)


davidtheclark:
Throwing this out in favor of removing `theme*` props, as described in https://github.com/mapbox/mr-ui/issues/26

PR #32, opened 2018-08-08
Popover changes

davidtheclark:
There are a lot of changes in this PR that are mostly just reformatting comments or improving documentation and, especially, documentation examples. But there are also some very important changes, which I hope are improvements, to Popover, PopoverTrigger, and Tooltip.

- There are 3 standard popover styles: light, dark, and warning. 
- `theme*` props have been replaced with higher-level styling props like `coloring` (light, dark, or warning) and `padded` (true or false).
- A new `passthroughProps` convention is introduced to pass props through a component to some child (most often to pass it through directly to a DOM element).
- Children of Tooltip no longer need to be functions!


davidtheclark:
@jseppi could you give this a review? There's a lot I think you could skip over, but if the changes and documentation make sense to you, and the examples (in docs and test cases) are clear and function as you'd expect, I think we'll be good to go.

jseppi:
👁‍🗨 Checking it out...

src/components/popover-trigger/popover-trigger.js
```diff
@@ -414,3 +383,93 @@ export default class PopoverTrigger extends React.Component {
...
+PopoverTrigger.propTypes = {
+  /**
+   * The trigger content. This can either be a string or valid JSX.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Accepts a CSS `display` value for the trigger: `'inline-block'` or
+   * `'block'`.
+   */
+  display: PropTypes.string,
```
jseppi: Maybe use `PropTypes.oneOf(['inline-block', 'block'])`

src/components/popover-trigger/popover-trigger.js
```diff
@@ -414,3 +383,93 @@ export default class PopoverTrigger extends React.Component {
...
+  respondsToFocus: PropTypes.bool,
+  /**
+   * Whether or not the trigger is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Props to pass directly to the [Popover](#popover) component. See Popover's
+   * documentation for details.
+   */
+  popoverProps: PropTypes.object,
```
jseppi: Maybe export Popover's propTypes and use it here?

src/components/popover-trigger/popover-trigger.js
davidtheclark: Ah, yeah, intended to do that.

src/components/popover-trigger/popover-trigger.js
davidtheclark: I worry that that might make it harder for babel-plugin-remove-prop-types to remove them, and I think the runtime check will work either way.

PR #36, opened 2018-08-11
Docs improvements

davidtheclark:
- Made a nicer syntax highlighting scheme.
- Fixed path names to components in examples.
- Fixed some links and small details in docs.
- Updated the Assembly version.
- Added a `<meta name="robots" content="noindex" />` tag, since once we publish this page on gh-pages we don't really want this page getting indexed — at least not yet, since many examples in the docs are still very sloppy.

@danswick for review, please. I'm thinking that after this merges we could cut a release with all the pending changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ UI components for Mapbox projects.
 npm install @mapbox/mr-ui
 ```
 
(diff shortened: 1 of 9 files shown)


davidtheclark:
Thanks, @danswick. I do think that another PR has addressed that console error, but I'll clean it up if not.

Issue #52, opened 2018-11-06
NewTabLink component


lshig:
This is a VERY minor wrapper component we use in Account Dashboard and other Accounts related web apps. It's helpful in remembering to add some extra attributes for creating a new tab link. This may find itself useful as it relates to https://github.com/mapbox/mr-ui/issues/13#issuecomment-436012330.

```
<a target="_blank" rel="noopener" {...this.props}>
  {this.props.children}
</a>
```

Issue #13, opened 2018-07-27
GoForward and GoBack components


davidtheclark:
We have some unnecessary variation in links/buttons where the intention is to make you feel like you're going *forward* or *back*. This would be a super simple component, but probably worth standardizing.

In most places we use a right chevron after the text to indicate "Take me to the next thing". In some places we use a right arrow. I don't know of very many places where we have the "Take me back" action, but I think when we do we use an arrow. Sometimes the text is bold, sometimes it's not.

Here's what I propose:

- We use a chevron for both forward and back, since that's the most widespread usage right now, suggesting that it's favored by the most people.
- By default we bold the text so it looks & feels like a [text button](https://github.com/mapbox/mr-ui/issues/11). I guess we could make it unboldable, too, with a prop.
- Small and large sizes.

davidtheclark:
To be clear, I have almost no feelings about this, just want a standard version that we can use by default.

lshig:
👋 I'm going to pick this up.

If there's no objection, I'm going to update `ChevronousText` before creating a `GoLink`. (It's open for a better name.)

I think it'd be valuable for `ChevronousText` to have a boolean `iconBefore` prop but by default have a `chevron-right` icon. In the `GoLink` component, I plan to use the updated `ChevronousText` component and give it the option to be bolded/unbolded or dark/light. This is also open for discussion, but I see `GoLink` wrapped by an `<a />` tag, which may take the user straight to the URL or open a new tab.

This is part of the Accounts team's effort to consolidate our components and abstract them out to the shared libraries of components. For example we're looking to switch out the following `Forward` and `Back` components:

![screen shot 2018-11-05 at 11 48 03 am](https://user-images.githubusercontent.com/9087698/48022771-e3989280-e0f0-11e8-9f97-a42ea22072e5.png)
![screen shot 2018-11-05 at 11 48 57 am](https://user-images.githubusercontent.com/9087698/48022776-e5faec80-e0f0-11e8-8fb1-935809a632d4.png)

cc: @angel

davidtheclark:
`ChevronousText` does only one thing, really: it puts the icon after text and connects the icon to the last word of the text so that there is no possibility the icon will end up orphaned on another line. So it's only particularly useful for text that might be multiline and will include icons inline with the text. I think that's not the case for `GoLink`s (those should live on their own, outside of paragraphs, and be single lines) — so maybe `IconText` is more useful precedent for `GoLink` (rather than `ChevronousText`)?

[...]

Issue #53, opened 2018-11-06
Greeting component


lshig:
In the Account Dashboard we have a simple `Greeting` component that's used on the landing page. This component takes in a username prop. Right now it's not too special besides saying `Welcome, {username}!`, but we're hopeful that one day this greeting could be smarter by language, the time of day, season, special event, etc. Right now we only plan to use it as a heading on pages, but it could be one day placed in body text or styled to different heading sizes for pages, cards, or modals.

![screen shot 2018-11-06 at 11 09 06 am](https://user-images.githubusercontent.com/9087698/48087560-7ea67080-e1b4-11e8-9a06-d2ddc6f77fc4.png)

cc: @angel

Issue #51, opened 2018-11-06
Card component


lshig:
The Account Dashboard and Atlas Account Dashboard use a card to display an overview/preview of information that can lead to dedicated pages with more detailed information. This `Card` component should be a wrapper component with children props, can be with or without a header, and can be with or without an actionable icon button in the top right corner. In the past, we've used the actionable icon button to close/remove the card or to collapse the card in view.

<img width="738" alt="screen shot 2018-11-05 at 3 59 59 pm" src="https://user-images.githubusercontent.com/9087698/48034448-4cddcd00-e114-11e8-82d7-e3e802c493ae.png">

We're open to discussion about this component! 

cc: @angel

PR #54, opened 2018-11-09
Update ChevronousText, Add GoLink and NewTabLink

lshig:
### Related issues

This PR closes https://github.com/mapbox/mr-ui/issues/13 and closes https://github.com/mapbox/mr-ui/issues/52 as adding `GoLink` (79d30c3) is dependent on adding `NewTabLink` (9059937) and updating `ChevronousText` (afb1f7a and 5103c7d).

### Description of changes

👋 Open to discussion/feedback!

**ChevronousText**
- Remove `themeIcon` for 2 reasons: it broke from the last update to `Icon` and based on how Mapbox design has traditionally used this component. The text and chevron always match in color and the text can vary in `txt-s`, `txt-m`, and `txt-l` in size, but the chevron stays the same size.
- Add `iconBefore` prop. At Mapbox we've only used `ChevronousText` for links so it's beneficial for `GoLink` to wrap the styling behaviors that `ChevronousText` offers.
- Possible future feature for this component? Allow the user to choose a `left`, `right`, `up`, or `down` chevron. Based on the current name of the component, it's not intuitive that they're only allowed to use a `left` or `right` chevron.

**NewTabLink**
- New component, examples, and tests that opinionatedly applies `rel="noopener"` to encourage more browser content-sharing security.

**GoLink**
- New component, examples, and tests that uses `ChevronousText` and `NewTabLink`. Allows opinionated variation in the following: 
  - text size (`small`, `medium`, `large`)
  - link color (`light`, `dark`)
  - bolded or unbolded text
  - whether this is a go back link (`chevron-left` before text) or a go forward link (`chevron-right` after text)
  - whether link should go directly to the `href` or open it in new tab

### Testing

- Created/updated documentation examples for each component. See the documentation Batfish site with `npm run start-docs`. Noting here that the chevron looks off in the live reload view, but otherwise normal in the built site version with `npm run build-docs` and `npm run serve-static-docs`.
- Created/updated Jest test cases and snapshots for each component. See `npm test` and the test cases app with `npm start`.

ChevronousText test cases
___
NewTabLink test cases
___
GoLink test cases

### Notes for the future

This is 1 of the series of new features (https://github.com/mapbox/mr-ui/issues/51, https://github.com/mapbox/mr-ui/issues/53, and https://github.com/mapbox/mr-ui/issues/50) I'd like to add before updating the `CHANGELOG.md` and releasing a new version of `mr-ui`.

Expect to add the following to the `CHANGELOG.md`
```
- [feature] Add **GoLink** component.
- [feature] Add **NewTabLink** component.
- **ChevronousText**
  - 🚨 [breaking change] Remove `themeIcon` prop.
  - [feature] Add `beforeIcon` prop.
```

### DIFF
--- a/src/components/chevronous-text/chevronous-text.js
+++ b/src/components/chevronous-text/chevronous-text.js
@@ -2,46 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Icon from '../icon';
 
-const ChevronAfterText = ({ text }) => {
-  const splitText = text.split(' ');
-  const lastWord = splitText.pop();
-  const textWithoutLastWord = splitText.join(' ');
-
-  return (
-    <span>
-      {textWithoutLastWord}{' '}
-      <span className="txt-nowrap">
-        {lastWord}
-        <Icon name="chevron-right" inline={true} />
-      </span>
-    </span>
-  );
-};
-
-ChevronAfterText.propTypes = {
-  text: PropTypes.string.isRequired
-};
-
-const ChevronBeforeText = ({ text }) => {
-  const splitText = text.split(' ');
-  const firstWord = splitText.shift();
-  const textWithoutFirstWord = splitText.join(' ');
-
-  return (
-    <span>
-      <span className="txt-nowrap">
-        <Icon name="chevron-left" inline={true} />
-        {firstWord}
-      </span>{' '}
-      {textWithoutFirstWord}
-    </span>
-  );
-};
-
-ChevronBeforeText.propTypes = {
-  text: PropTypes.string.isRequired
-};
-
 export default class ChevronousText extends React.PureComponent {
   static propTypes = {
     /** When true, the text will follow after a left pointed chevron. */
@@ -56,11 +16,26 @@ export default class ChevronousText extends React.PureComponent {
 
   render() {
     const { iconBefore, text } = this.props;
+    const splitText = text.split(' ');
+    const iconWord = iconBefore ? splitText.shift() : splitText.pop();
+    const textWithoutIconWord = splitText.join(' ');
 
     return iconBefore ? (
-      <ChevronBeforeText text={text} />
+      <span className="inline-block">
+        <span className="txt-nowrap">
+          <Icon name="chevron-left" inline={true} />
+          {iconWord}
+        </span>{' '}
+        {textWithoutIconWord}
+      </span>
     ) : (
-      <ChevronAfterText text={text} />
+      <span className="inline-block">
+        {textWithoutIconWord}{' '}
+        <span className="txt-nowrap">
+          {iconWord}
+          <Icon name="chevron-right" inline={true} />
+        </span>
+      </span>
     );
   }
 }
--- a/src/components/chevronous-text/chevronous-text.js
+++ b/src/components/chevronous-text/chevronous-text.js
@@ -6,16 +6,22 @@ export default class ChevronousText extends React.PureComponent {
   static propTypes = {
     /** When true, the text will follow after a left pointed chevron. */
     iconBefore: PropTypes.bool,
+    /**
+     * The width and height size of the chevron icon. Note that this icon is
+     * inline and the height won't go beyond the line-height.
+     */
+    iconSize: PropTypes.number,
     /** The text that should be aligned next to the chevron. */
     text: PropTypes.string.isRequired
   };
 
   static defaultProps = {
-    iconBefore: false
+    iconBefore: false,
+    iconSize: 18
   };
 
   render() {
-    const { iconBefore, text } = this.props;
+    const { iconBefore, iconSize, text } = this.props;
     const splitText = text.split(' ');
     const iconWord = iconBefore ? splitText.shift() : splitText.pop();
     const textWithoutIconWord = splitText.join(' ');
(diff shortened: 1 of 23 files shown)


src/components/go-link/go-link.js
```diff
@@ -0,0 +1,57 @@
...
+    return isNewTab ? (
+      <NewTabLink href={href} className={linkClasses}>
+        <ChevronousText text={text} iconBefore={goBack} />
+      </NewTabLink>
+    ) : (
+      <a href={href} className={linkClasses}>
+        <ChevronousText text={text} iconBefore={goBack} />
+      </a>
+    );
+  }
```
davidtheclark: Hm ... it's unclear to me what value this component provides.

- The `isNewTab` & `href` props might not be necessary because the user could use a wrapper `<NewTabLink>` or `<a>` directly.
- The `color`, `size`, and `isBold` props could be removed and the user could set those props on a wrapper element.

If ChevronousText already solves the two small problems for the component (positioning the chevron and handling multiline text), and your fix in this PR allows it to be used for a link to go back — do we need this additional component?

src/components/new-tab-link/new-tab-link.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This component applies an opinionated relationship (rel) attribute to
+ * prevent the new tab from accessing the originating tab's content.
+ */
+export default class NewTabLink extends React.Component {
+  static propTypes = {
+    /** The content of the new tab link. */
+    children: PropTypes.node.isRequired,
+    /** The new tab location path. */
+    href: PropTypes.string
```
davidtheclark: Should this prop be required? The link would have no purpose without it, right?

src/components/chevronous-text/chevronous-text.js
```diff
@@ -4,27 +4,36 @@ import Icon from '../icon';
...
-        {textWithoutLastWord}{' '}
+    return iconBefore ? (
+      <span className="inline-block">
+        <span className="txt-nowrap">
+          <Icon name="chevron-left" inline={true} />
+          {iconWord}
+        </span>{' '}
+        {textWithoutIconWord}
+      </span>
+    ) : (
```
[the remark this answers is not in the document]
lshig: :+1: Added this change in ddce256. Did the same for `GoLink` in 8e810e4.

src/components/new-tab-link/new-tab-link.js
lshig: :+1: You're right. `href` should be required and shouldn't be `""` by default. In that scenario, clicking on the link would open a new tab with the same location as where it originated from (like how Duplicate works). If a user wants that functionality, they should explicitly add it themselves. Added this change in 7a02173.

src/components/go-link/go-link.js
lshig: This component is limited in styling options and has a default to fall back on if links need to be added quickly. For Accounts related web apps, we either re-implement this component or copy/paste the same few lines of code as needed throughout the apps. Unfortunately that makes it very easy for unintended variation in design. It'd be ideal for us to consolidate and share the same design standards for all of our links. If someone wants more flexibility than what is standardized here, they should probably continue to create their own components or create a wrapper components around `ChevronousText`. Alternatively the Accounts team is open to adding this more opinionated/limited component to our own shared components repo similar to the Documentation team.

lshig:
> Did you intend to keep this behavior (it looks like it's still working this way), or did you consider this behavior a bug? (I might lean towards considering it a bug, since it doesn't seem natural that the chevron would be the same size for very small and very large text.)

Originally this was a minor breaking change from `Icon` not having `iconTheme`. And then... I kept the behavior because there wasn't any Mapbox use cases of the text being so large that the chevron looked extra small (besides the `ChrevonousText` example in the documentation). Design-wise, it makes sense to at least allow the option to adjust the icon size in comparison to the text. I added this feature in 8e4d9f4 by adding an `iconSize` prop in `ChevronousText`. It will continue to allow a wrapper component around `ChevronousText` to determine if the icon size grows or stays at `18` by default. `Icon`'s `size` prop is also set to `18` by default.

> Do you know why the instance of `ChevronousText` in the documentation look misaligned?

The built version `lineHeight` from `setHeight()` in `Icon` returns back `24px`, while the development live server version returns back `normal`. Therefore... it won't adjust the height of the inlined `Icon`. I updated the example to ensure that a line-height was provided in 8e4d9f4, similar to `Icon`'s '`icon-b.js` example.

src/components/chevronous-text/chevronous-text.js
```diff
@@ -4,27 +4,46 @@ import Icon from '../icon';
...
     /** The text that should be aligned next to the chevron. */
-    text: PropTypes.string.isRequired,
-    /** Additional classes to apply to the icon element */
-    themeIcon: PropTypes.string
+    text: PropTypes.string.isRequired
+  };
+
+  static defaultProps = {
+    iconBefore: false,
+    iconSize: 18
```
davidtheclark: I think this prop is a good solution 👍 I wonder, though, if the default value should be `1.5em` or `1.6em` to approximately match line-height by default, instead of being a hard-coded `18`? (I got those `em` ratios by reviewing the Assembly line-height/font-size ratios for the `txt-{size}` classes.) I'd imagine that in the default cause we'd want a smaller chevron if we had smaller text?

Would that also alleviate the need for special icon-size handling in GoLink?

davidtheclark: Fantastic 🎿 🐻

PR #55, opened 2018-11-15
Port from Batfish to Underreact

danswick:
See https://github.com/mapbox/frontend-platform-team/issues/32. For sites like this one, which do not need to generate multiple HTML files, @mapbox/frontend-platform has a new favored tool: Underreact. It's significantly simpler than Batfish, and is using the latest, fastest versions of Webpack and Babel — so it's the better choice for any sites that don't need Batfish's HTML generation.

A few notes about the changes in this PR:

- I used `<link>` and `<script>` tags in the HTML to load mbx-assembly.
- The new `app.js` component is just the content of the old `src/docs/app.js`.
- `browserslist: ['defaults']` means Underreact won't bother building for IE11, because this is an internal site (that should speed things up).
- Babel did not like the `Object:` type [here](https://github.com/mapbox/mr-ui/compare/underreact?expand=1#diff-eeb6dd784bbd7ad8087727499500dffbL28) and I'm not familiar enough with Babel to configure my way around that issue, so I just removed it. 

Next steps:
- [x] Get the page to render 😬 
- [x] Update tests

@davidtheclark: I would love to talk through this a bit with you tomorrow. I am a bit fuzzy on what's going on with the Babel 7 updates I made here and I am still not entirely sure why the page isn't being rendered.

### DIFF
--- a/.artifacts.yml
+++ b/.artifacts.yml
@@ -1,3 +1,0 @@
-version: v1
-defaults:
-  - publisher
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,5 +6,5 @@ src/docs/data
 pkg
 src/test-cases-app/component-index.js
 dist
-*underreact.config.js
+**/*underreact.config.js
 html.js
--- a/scripts/build-module-indexes.js
+++ b/scripts/build-module-indexes.js
@@ -5,8 +5,9 @@ require('hard-rejection/register');
 const fs = require('fs');
 const path = require('path');
 const pify = require('pify');
+
 const excludeDirs = new Set(['utils']);
-const srcRoot = path.resolve(__dirname, '../src/components/');
+const srcRoot = path.resolve(__dirname, '../src/components');
 
 function generateIndex(moduleName) {
   const content = `import main from './${moduleName}';
--- a/src/docs/docs-underreact.config.js
+++ b/src/docs/docs-underreact.config.js
@@ -3,7 +3,7 @@ const path = require('path');
 
 module.exports = () => {
   return {
-    siteBasePath: '/mr-ui/docs',
+    siteBasePath: '/mr-ui',
     browserslist: ['defaults'],
     jsEntry: path.resolve(__dirname, './index.js'),
     htmlSource: html
--- a/src/test-cases-app/html.js
+++ b/src/test-cases-app/html.js
@@ -5,8 +5,8 @@ module.exports = `
     <title>Test cases</title>
     <meta charset='utf-8'>
     <meta name='viewport' content='width=device-width, initial-scale=1'>
-    <link href="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.min.css" rel="stylesheet">
-    <script async src="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.js"></script>
+    <link href="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.1/assembly.min.css" rel="stylesheet">
+    <script async src="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.1/assembly.js"></script>
   </head>
   <body>
   </body>
--- a/src/docs/html.js
+++ b/src/docs/html.js
@@ -7,12 +7,12 @@ module.exports = `
     <meta name="robots" content="noindex" />
     <title>Mr. UI</title>
     <link
-      href="https://api.mapbox.com/mapbox-assembly/mbx/v0.27.0/assembly.min.css"
+      href="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.1/assembly.min.css"
       rel="stylesheet"
     />
     <script
       async
-      src="https://api.mapbox.com/mapbox-assembly/mbx/v0.27.0/assembly.js"
+      src="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.1/assembly.js"
     ></script>
   </head>
   <body>
--- a/src/test-cases-app/test-cases-app-underreact.config.js
+++ b/src/test-cases-app/test-cases-app-underreact.config.js
@@ -3,7 +3,7 @@ const path = require('path');
 
 module.exports = () => {
   return {
-    siteBasePath: '/',
+    siteBasePath: '/', // react-test-kitchen does not currently support configurable basepaths
     jsEntry: path.resolve(__dirname, './index.js'),
     htmlSource: html
   };


danswick:
Ok, after much flailing, I think this is ready to have its WIP tag removed. Tests are passing** and the page is rendering. 

** tests are passing _locally_, but Travis is failing because it can't find `@mapbox/underreact`. @kepta: when I try to go to `https://www.npmjs.com/login?next=/package/@mapbox/underreact/v/0.1.5` (or similar), I'm prompted to log in, which makes me think the package has recently become private.

kepta:
> ** tests are passing locally, but Travis is failing because it can't find @mapbox/underreact. @kepta: when I try to go to https://www.npmjs.com/login?next=/package/@mapbox/underreact/v/0.1.5 (or similar), I'm prompted to log in, which makes me think the package has recently become private.

Looks like I have been mistakingly publishing it as private. I have publicly published it now and things should be working fine. Sorry about that.

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 coverage
 .DS_Store
 _batfish*
+_underreact-site
```
davidtheclark: Let's upgrade to the latest version of Underreact, in which case this will be `_site`.

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 coverage
 .DS_Store
 _batfish*
```
davidtheclark: Can we delete this line?

src/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
davidtheclark: If this file is specific to docs, lets put it in `src/docs`.

underreact.config.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
davidtheclark: Is this for the docs or the test cases? Let's run both with Underreact.

kepta:
Looks like this might be a great place to enable hot reloading feature of Underreact 🔥. We can either add it in this PR or I can send another PR to specifically enable it.

underreact.config.js
davidtheclark: Have you looked at how each app works now, without Underreact? That would be the place to start.

Underreact commands have a `--config` option that you can use in order to store two different configs in one repo and run them in different Underreact processes.

kepta: To give an example here is a package.json using an explicit config:

https://github.com/mapbox/streets-review-frontend/blob/fb35b1ad05da326ee20a125e6599569c2f7d0cf2/package.json#L27

danswick:
I've done some reorganizing here that I hope will make the project easier to maintain: 

- Moved docs and the test-cases-app into their own directories separate from the source components. 
- Finished scaffolding for the docs page's port to underreact.
- Begain porting the test-cases-app to underreact, but hit a dead-end trying to get it working (and also got myself all kinds of turned around trying to get eslint configured). Underreact seems to be building the page without issue and my React chrome dev tools indicate that the test cases component and ReactTestKitchen component are being created, but they're not being populated with anything and are not throwing any errors. I'm a bit lost on where things are going wrong and could use some backup if anyone from @mapbox/frontend-platform is wiling to take a look with me.

kepta:
@danswick this was an interesting bug. I found that the bug is with the way we are using the basePath as `'/mr-ui/test-cases-app'`, but the `react-test-kitchen` expects the basePath to be `/` and that's the client side routing in `rtk` was wasn't outputting anything.

https://github.com/mapbox/react-test-kitchen/blob/1dbc8fb4e33c86cd02b66bd3f077587abe287a42/src/react-test-kitchen.js#L19-L23

We can probably send a fix to `rtk` to understand base paths or you can simply use `/` to keep it happy 😅.

Finding this bug was pretty confusing. I found the way this repo auto-generates the files pretty confusing (Maybe it is just me being unfamiliar with the `rtk` testing pattern). Also, I wasn't careful to see that budo server is also being used and the fact that you reorganised the repo structure made the auto scripts generation by `scripts/` broken. 

If it is not too much of a trouble, I suggest save the repo structure change to a separate issue/PR. This would make it easier to grok at the diff generated in this PR and also keep changes just related to porting of Underreact.

davidtheclark:
> I suggest saving the repo structure change to a separate issue/PR

Yes, definitely.

davidtheclark:
> I found it confusing to have two web apps that reference the component source, but the component source lives alongside just one of the apps

I don't think I understand the confusion here. We have `src/components`, `src/docs`, and `src/test-cases`, so the source components are already alongside both apps.

danswick:
Thanks for all of the feedback, @kepta and @davidtheclark. I went ahead and fixed up the project structure, swapped the test cases app to a `/` basepath (thanks for that hint, @kepta! side note: I started experimenting with configurable basepath in `rtk` [here](https://github.com/mapbox/react-test-kitchen/tree/basePath), but was having some trouble and opted to stick with a basepath of `/` since the test cases app is really only ever used locally), and updated snapshots to get tests passing. Would love another review now that the diff is more manageable 🙏

.artifacts.yml
```diff
@@ -0,0 +1,3 @@
+version: v1
+defaults:
+  - publisher
```
davidtheclark: Mr UI isn't currently being published with Publisher, right?

.eslintignore
```diff
@@ -6,3 +6,5 @@ src/docs/data
 pkg
 src/test-cases-app/component-index.js
 dist
+*underreact.config.js
```
davidtheclark: Should this be `**/underreact.config.js`?

.eslintignore
```diff
@@ -6,3 +6,5 @@ src/docs/data
...
+html.js
```
davidtheclark: Hm, we shouldn't need to do this. But I don't want to clog up the PR for linting. Let's sort it out after the merge. (Hint: I think we'll use ESLint's `overrides` option.)

scripts/build-module-indexes.js
```diff
@@ -5,9 +5,8 @@ require('hard-rejection/register');
 const fs = require('fs');
 const path = require('path');
 const pify = require('pify');
-
```
davidtheclark: 👹

scripts/build-module-indexes.js
```diff
@@ -5,9 +5,8 @@ require('hard-rejection/register');
...
 const excludeDirs = new Set(['utils']);
-const srcRoot = path.resolve(__dirname, '../src/components');
+const srcRoot = path.resolve(__dirname, '../src/components/');
```
davidtheclark: 👹

(This change shouldn't make any difference, right?)

src/docs/docs-underreact.config.js
```diff
@@ -0,0 +1,11 @@
+const html = require('./html.js');
+const path = require('path');
+
+module.exports = () => {
+  return {
+    siteBasePath: '/mr-ui/docs',
```
davidtheclark: The true base path is just `/mr-ui`, right? https://mapbox.github.io/mr-ui/

src/test-cases-app/html.js
```diff
@@ -0,0 +1,14 @@
+module.exports = `
+  <!DOCTYPE html>
+  <html lang='en'>
+  <head>
+    <title>Test cases</title>
+    <meta charset='utf-8'>
+    <meta name='viewport' content='width=device-width, initial-scale=1'>
+    <link href="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.min.css" rel="stylesheet">
+    <script async src="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.js"></script>
```
davidtheclark: 🤔 it's suspicious to me that we have different version of Assembly in the test cases and docs. Are you copying a mistake already in place, or is this a typo?

.artifacts.yml
danswick: Oh yowza. I just got into Underreact robot mode and put this here automatically. I can remove.

.eslintignore
danswick: It works this way, but `**/*underreact.config.js` is more precise.

.eslintignore
danswick: Sounds good 👍

scripts/build-module-indexes.js
danswick: 👺

src/docs/docs-underreact.config.js
danswick: Ah, yep! This is a gremlin from when I was trying to give `docs` and `test-cases-app` their own base paths under `mr-ui` 😬

src/test-cases-app/html.js
danswick: Looks like it was [already](https://github.com/mapbox/mr-ui/blob/master/src/test-cases-app/index.html) in [place](https://github.com/mapbox/mr-ui/blob/master/src/docs/pages/index.js), but I just bumped them both up to `0.28.1` and tested locally and everything looks good 👍

Issue #50, opened 2018-11-05
Alert component


lshig:
For the Account Dashboard and other Accounts team related web apps, we have a localized Alert component that appears when there's a confirmation, warning, or error on modals and cards.

<img width="549" alt="screen shot 2018-11-05 at 3 38 16 pm" src="https://user-images.githubusercontent.com/9087698/48033792-9e388d00-e111-11e8-8266-da3904d1b750.png">

<img width="767" alt="screen shot 2018-11-05 at 3 42 55 pm" src="https://user-images.githubusercontent.com/9087698/48033793-a1337d80-e111-11e8-8307-10df4408c8fe.png">

It can also be globalized with a wrapper that allows the user to close it early.

<img width="272" alt="screen shot 2018-11-05 at 3 42 18 pm" src="https://user-images.githubusercontent.com/9087698/48033879-05564180-e112-11e8-968f-6ddb4232ed9d.png">

Open to any discussion about this component!

cc: @angel

PR #59, opened 2018-11-21
Add ControlAlert component, tests, and examples

lshig:
### Related issues

This PR closes https://github.com/mapbox/mr-ui/issues/50 as it adds `ControlAlert`.

### Description of changes

👋 Open to discussion/feedback!

**ControlAlert**

This kind of alert has been used by Accounts-related web apps.

- New component, examples, and tests that uses `Icon` and `Tooltip`. Allows opinionated variation in the following:
  - autoFocus
  - onButtonClick
  - theme ('error', 'locked', 'success', 'warning')

- Conscious design decisions with @angel:
  - While cards and modals can `Close` or `Collapse`, alerts are `Dismiss`ed
  - Theme coloring and icons are set by severity:
    - success (green)
    - warning or locked (yellow)
    - error (red)
  - TODO:
    - Should the icons scale with the children text? We do this for `ChevronousText` but currently not for `ControlCard`. The screenshots below display `Icon` with a set width of `18px` and an adjusting inline height to the the parent element's line-height.

### Testing

- Created/updated documentation examples for each component. See the documentation Batfish site with `npm run start-docs`.
- Created/updated Jest test cases and snapshots for each component. See `npm test` and the test cases app with `npm start`.

ControlAlert doc examples

ControlAlert test cases

### DIFF
--- a/src/components/control-alert/control-alert.js
+++ b/src/components/control-alert/control-alert.js
@@ -1,0 +1,98 @@
+import classnames from 'classnames';
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../icon';
+import Tooltip from '../tooltip';
+
+export default class ControlAlert extends React.Component {
+  static propTypes = {
+    /** Whether to use ARIA alert role for screen readers or not. */
+    autoFocus: PropTypes.bool,
+    /** The content of the alert. */
+    children: PropTypes.node.isRequired,
+    /** Called on click of dismiss button. */
+    onButtonClick: PropTypes.func,
+    /**
+     * The alert `theme` options are "error", "locked", "success", and
+     * "warning".
+     */
+    theme: PropTypes.oneOf(['error', 'locked', 'success', 'warning']).isRequired
+  };
+
+  static defaultProps = {
+    autoFocus: false
+  };
+
+  componentDidMount() {
+    const { autoFocus } = this.props;
+
+    if (autoFocus && this.contentRef) {
+      this.contentRef.focus();
+    }
+  }
+
+  handleContentRef = ref => {
+    this.contentRef = ref;
+  };
+
+  onButtonClick = () => {
+    const { onButtonClick } = this.props;
+
+    if (onButtonClick) {
+      onButtonClick();
+    }
+  };
+
+  renderDismissButton() {
+    const { onButtonClick } = this.props;
+
+    if (!onButtonClick) return null;
+
+    return (
+      <div>
+        <Tooltip content="Dismiss" block={true}>
+          <button
+            className="bg-transparent color-gray-dark color-blue-on-hover px0 py0"
+            data-test="alert-dismiss"
+            onClick={this.onButtonClick}
+          >
+            <Icon name="close" inline={true} />
+          </button>
+        </Tooltip>
+      </div>
+    );
+  }
+
+  render() {
+    const { autoFocus, children, theme } = this.props;
+    const containerClasses = classnames('round px24 py24 flex-parent', {
+      'bg-green-light': theme === 'success',
+      'bg-red-faint': theme === 'error',
+      'bg-yellow-light': theme === 'warning' || theme === 'locked'
+    });
+
+    let iconName = 'alert';
+    if (theme === 'success') {
+      iconName = 'check';
+    } else if (theme === 'locked') {
+      iconName = 'lock';
+    }
+
+    return (
+      <div className={containerClasses}>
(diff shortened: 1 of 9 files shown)


angel:
@lshig these look good to me, why is the yellow power ranger centered and in caps?

lshig:
@angel Oh. Just an example of what's possible... Yellow, eye-catching, and right in the center.

yellow power ranger

The documentation examples will be hosted at https://mapbox.github.io/mr-ui/ once these changes are released. They can also be viewable locally if you run `npm run start-docs`.

angel:
@lshig can we restrict that the text is always left aligned?

lshig:
@angel We can make it a general rule of thumb by design: when we use this component in web apps at Mapbox, we should by default let `ControlAlert`'s children body text be left aligned and sentence cased. I don't think it should be the responsibility of `ControlAlert` to enforce the prop `children`'s styling since we vary in use of margins, padding, capitalization, text size, text alignment, etc.

We discussed this on Slack, but FYI to anyone curious, we decided to scale the `ControlAlert` icons with `1.3em` width. Also to encourage text to be sentence cased or left aligned by example, I adjusted the doc example at 9e44044.

screen shot 2018-11-21 at 11 41 41 am

src/components/control-alert/control-alert.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+  };
+
+  renderDismissButton() {
+    const { onButtonClick } = this.props;
+
+    if (!onButtonClick) return null;
+
+    return (
+      <div>
```
davidtheclark: Any need for this wrapper div?

src/components/control-alert/control-alert.js
```diff
@@ -0,0 +1,98 @@
...
+import Icon from '../icon';
+import Tooltip from '../tooltip';
+
+export default class ControlAlert extends React.Component {
+  static propTypes = {
+    /** Whether to use ARIA alert role for screen readers or not. */
+    autoFocus: PropTypes.bool,
+    /** The content of the alert. */
+    children: PropTypes.node.isRequired,
+    /** Called on click of dismiss button. */
```
davidtheclark: We should document that the button will only appear if you've added this callback.

src/components/control-alert/examples/control-alert-example-options.js
```diff
@@ -0,0 +1,66 @@
+/*
+Controlled alert with options.
```
davidtheclark: I'm not sure "Controlled" is descriptive enough here.

src/components/control-alert/control-alert.js
```diff
@@ -0,0 +1,98 @@
...
+        </div>
+        <div
+          className="flex-child flex-child--grow"
+          ref={this.handleContentRef}
+          role={autoFocus ? 'alert' : null}
+          tabIndex={autoFocus ? -1 : null}
+        >
+          {children}
+        </div>
+        {this.renderDismissButton()}
```
davidtheclark: In my opinion, the positioning of this button is a little unusual: it's further away from the edges of the container than you'd typically see with a "close" button like this, and that distance is especially striking when the text is multiline. But if @angel is cool with it, 👍 from me.

angel: I'm good with the placement of the x! Maybe we can add a little more padding between the text and the x so there's more breathing room for the x. Also, now that I look at it again, I wonder if it can be a little bigger so it's the same height of the icon on the right. Right now it looks a little smaller. So it can stand out a little more for multiline text.

src/components/control-alert/control-alert.js
lshig: 👍Added `Must have this callback to display the button.` in 64a4431.

src/components/control-alert/examples/control-alert-example-options.js
lshig: 👍Documentation can always be improved. I updated the `ControlAlert` and `ControlCard` with options example descriptions in 56cd281. We may want to update all the other vague component example descriptions at some point.

src/components/control-alert/control-alert.js
lshig: > ... now that I look at it again, I wonder if it can be a little bigger so it's the same height of the icon on the right. ...

@angel We may have to revisit how the icon svgs are sized and scaled in Assembly. They are located [here](https://github.com/mapbox/assembly/tree/mb-pages/src/svgs). In the multiline error alert picture, the `error` and `close` icons are both at `1.3em` width and `24px` height.

PR #60, opened 2018-11-30
PageHeader component draft

davidtheclark:
Draft component for a shareable page header. I want to confirm the appearance with anyone interested before finishing it off with tests and documentation. I'm imagining this will be used in apps owned by @mapbox/studio and @mapbox/accounts. **Please share your ideas about how this looks and works**, here or in chat!

Here's what you get:

- Non-mobile
  - Mapbox logo on the left linking to the homepage.
  - Just to the right of that, the name of the app, linking to the app subdomain home (`href="/"`).
  - Single-level navigation (i.e. no dropdowns).
  - Space on the far right for us to insert the user menu trigger.
- Mobile
  - App nav trigger on the left, Mapbox mark and app name between, user menu trigger on the right.
  - App nav opens in a panel on the left. Has the Mapbox mark and app name at the top, links beneath, close button on the right.
  - When app nav is open, there's a dark background over the page, and when you touch it the nav closes.
- Mobile header shows when you're under the `-mm` breakpoint, non-mobile header shows above that.

Pictures:

screen shot 2018-11-30 at 11 46 09 am
screen shot 2018-11-30 at 11 46 39 am
screen shot 2018-11-30 at 11 46 47 am

cc @mapbox/frontend @mapbox/frontend-platform

### DIFF
--- a/src/components/page-header/__tests__/page-header-test-cases.js
+++ b/src/components/page-header/__tests__/page-header-test-cases.js
@@ -24,7 +24,7 @@ testCases.basic = {
   }
 };
 
-testCases.accounts = {
+testCases.account = {
   description: 'Account',
   component: PageHeader,
   props: {
--- a/src/components/page-header/__tests__/page-header.test.js
+++ b/src/components/page-header/__tests__/page-header.test.js
@@ -12,19 +12,17 @@ let wrapper;
 describe(testCases.basic.description, () => {
   beforeEach(() => {
     testCase = testCases.basic;
+    wrapper = shallow(React.createElement(testCase.component, testCase.props));
   });
 
   test('renders as expected', () => {
-    const wrapper = shallow(
-      React.createElement(testCase.component, testCase.props)
-    );
     expect(wrapper).toMatchSnapshot();
   });
 });
 
-describe(testCases.accounts.description, () => {
+describe(testCases.account.description, () => {
   beforeEach(() => {
-    testCase = testCases.accounts;
+    testCase = testCases.account;
     wrapper = shallow(React.createElement(testCase.component, testCase.props));
   });
 
--- a/src/components/page-header/page-header.js
+++ b/src/components/page-header/page-header.js
@@ -5,6 +5,8 @@ import Icon from '../icon';
 const MOBILE_NAV_TRIGGER_WIDTH = 36;
 const MOBILE_NAV_POINTER_WIDTH = 32;
 const MOBILE_NAV_MENU_OFFSET_LEFT = 10;
+const MBX_USER_MENU_WIDTH = 66;
+const MOBILE_NAV_MENU_OFFSET_TOP = 56;
 
 class NonMobilePageHeader extends React.Component {
   shouldComponentUpdate() {
@@ -30,30 +32,30 @@ class NonMobilePageHeader extends React.Component {
       <div className="limiter flex-parent flex-parent--center-cross">
         <div className="flex-child flex-child--no-shrink flex-parent flex-parent--center-cross">
           <a
-            className="flex-child mb-logo wmax180-mxl wmax30-mm"
+            className="flex-child mb-logo wmax180-mxl wmax30"
             href="https://www.mapbox.com"
             aria-label="Mapbox"
           />
           <div className="flex-child flex-parent flex-parent--center-cross color-blue txt-l txt-bold relative">
             <div
-              className="flex-child mx12 bg-blue h24 none-mm block-mxl"
+              className="flex-child mx12 bg-blue h24 none block-mxl"
               style={{ width: 2 }}
             />
             <a
               href="/"
-              className="flex-child color-blue color-blue-dark-on-hover ml6-mm ml0-mxl"
+              className="flex-child color-blue color-blue-dark-on-hover ml6 ml0-mxl"
             >
               {siteName}
             </a>
           </div>
         </div>
 
-        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s mt6">
+        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s mt3">
           {itemEls}
         </nav>
         <div
           id="mbx-user-menu"
-          style={{ width: 66 }}
+          style={{ width: MBX_USER_MENU_WIDTH }}
           className="flex-child flex-child--no-shrink"
         />
       </div>
@@ -76,20 +78,11 @@ class MobilePageHeader extends React.Component {
     return this.state.open !== nextState.open;
   }
 
-  componentDidMount() {
-    const doc = window.document;
-    doc.addEventListener('click', this.closeOnDocumentClick);
-  }
-
-  componentWillUnmount() {
-    const doc = window.document;
-    doc.removeEventListener('click', this.closeOnDocumentClick);
-  }
-
   closeOnDocumentClick = event => {
     // Close the popup if the target is outside of the popup
     if (!this.state.open) return;
-    if (document.getElementById('mobile-menu-trigger').contains(event.target)) {
+    const mobileMenuTrigger = this.mobileTriggerRef;
+    if (mobileMenuTrigger.contains(event.target)) {
       return;
     }
     if (
@@ -102,7 +95,13 @@ class MobilePageHeader extends React.Component {
   };
 
   toggleMenu = () => {
-    this.setState(state => ({ open: !state.open }));
+    if (this.state.open) {
+      this.setState(() => ({ open: false }));
+      const doc = window.document;
+      doc.removeEventListener('click', this.closeOnDocumentClick);
+      return;
+    }
+    this.setState(() => ({ open: true }));
   };
 
   renderLogoName() {
@@ -129,6 +128,9 @@ class MobilePageHeader extends React.Component {
 
   renderMenu() {
     if (!this.state.open) return null;
+
+    const doc = window.document;
+    doc.addEventListener('click', this.closeOnDocumentClick);
 
     const pointerStyle = {
       width: MOBILE_NAV_POINTER_WIDTH / 2,
@@ -167,18 +169,16 @@ class MobilePageHeader extends React.Component {
 
     return (
       <div
-        id="mobile-menu-container"
-        className="absolute left shadow-darken10-bold bg-white round"
-        style={{ top: '56px', marginLeft: MOBILE_NAV_MENU_OFFSET_LEFT }}
+        className="absolute left shadow-darken10-bold bg-white round z5"
+        style={{
+          top: MOBILE_NAV_MENU_OFFSET_TOP,
+          marginLeft: MOBILE_NAV_MENU_OFFSET_LEFT
+        }}
         ref={node => {
           this.menuBodyElement = node;
         }}
       >
-        <div
-          id="mobile-menu-pointer"
-          className="color-white z5"
-          style={pointerStyle}
-        />
+        <div className="color-white" style={pointerStyle} />
         <div className="py30 px30">
           <nav>
             <ul className="txt-bold color-darken75">{itemEls}</ul>
@@ -194,8 +194,10 @@ class MobilePageHeader extends React.Component {
         <div className="limiter flex-parent flex-parent--center-cross">
           <div className="flex-child flex-child--no-shrink">
             <button
-              id="mobile-menu-trigger"
               type="button"
+              ref={el => {
+                this.mobileTriggerRef = el;
+              }}
               aria-label="menu"
               className="block color-blue"
               onClick={this.toggleMenu}
@@ -211,7 +213,7 @@ class MobilePageHeader extends React.Component {
           </div>
           <div
             id="mbx-user-menu-mobile"
-            style={{ width: 66 }}
+            style={{ width: MBX_USER_MENU_WIDTH }}
             className="flex-child flex-child--no-shrink"
           />
         </div>
--- a/src/test-cases-app/index.html
+++ b/src/test-cases-app/index.html
@@ -4,8 +4,8 @@
   <title>Test cases</title>
   <meta charset='utf-8'>
   <meta name='viewport' content='width=device-width, initial-scale=1'>
-  <link href="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.min.css" rel="stylesheet">
-  <script async src="https://api.mapbox.com/mapbox-assembly/mbx/v0.26.0/assembly.js"></script>
+  <link href="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.2/assembly.min.css" rel="stylesheet">
+  <script async src="https://api.mapbox.com/mapbox-assembly/mbx/v0.28.2/assembly.js"></script>
 </head>
 <body>
   <div id="app"></div>


davidtheclark:
cc @mapbox/docs to take a look, for ideas, etc., though you'll need a different implementation.

elifitch:
More of a design concern, but I'd love it if the hamburger menu button and the close menu button were in the same spot. Makes it a little more convenient to use.

davidtheclark:
@elifitch Thanks for chiming in. We can do that. My concern is that it then puts the logo + sitename in a weird place on the right. Suggestion: if we want the close button on the left, we make the menu open up to the full width of the viewport, extending down only as far as it needs to go. What do you think of that?

elifitch:
> OR we have the slider slide out beneath the header, instead of on top of it.

This seems like a nice compromise :)

lshig:
> OR we have the slider slide out beneath the header, instead of on top of it.

In doing this, do you imagine that the hamburger menu icon will animate or fade into the `x` icon? I like how our current mobile menu keeps the Mapbox logo and the icon in the same place.

cc: @angel

lshig:
Also, question related to the `User` placeholder, will the user menu also have a mobile version slider or will it remain a dropdown type menu in desktop and mobile screens? If we do the slider as mentioned previously, I'm having a hard time imagining how those 2 menus will compete with each other for space. Will they overlap? Will one animate in and the other animate out of view?

angel:
Could we have the user be a part of the menu then the menu can be back on the right hand side and the logo can stay on the left, and nothing has to move?

davidtheclark:
Initial version will have no animation, unless @mapbox/accounts or @mapbox/studio has the opportunity to contribute that feature. Animation can always be added later.

User menu will be a dropdown, as it is now, rather than a (non-sliding) slider. This way it can work essentially the same on desktop and mobile.

Idea: the app menu could also be a dropdown. That could provide some nice symmetry between the two menus.

If you click outside one of the menus, it should close. So if you have the user menu open and you click the app menu trigger, the user menu should close.

> Could we have the user be a part of the menu then the menu can be back on the right hand side and the logo can stay on the left, and nothing has to move?

I don't think so, unless we were to scratch the user menu idea. The user menu is a standalone piece that @mapbox/frontend-platform will maintain. I don't know of a technically reasonable way for us to blend that into the subdomain-specific menus for apps and docs.

danswick:
I made some tweaks based on the discussion in this ticket and would love feedback from folks. The biggest change was moving to a simple dropdown from a non-sliding slider per https://github.com/mapbox/mr-ui/pull/60#issuecomment-443908112. I tried to make it look as much like the user menu as possible without going full pointy-popover and I'm not sure if it works well or if it's weird. 

image

davidtheclark:
@danswick Positioning looks good. I think it needs more work to match the existing dropdowns. I can tell it lacks the triangle and the padding of the existing dropdowns. Maybe that's it?

danswick:
Thanks for the feedback, @davidtheclark. The latest commit adds a dynamically-positioned pointer triangle, updates padding to match that of the other dropdowns in the existing page shell, and cleans up some CSS nonsense from my previous commit. 

image

danswick:
image

image


Pointer and spacing should be pretty close now, @davidtheclark. For the vertical position, I made a temporary change to initialize the user menu and compared the two.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+          className="flex-child relative mr18 mr36-mxl color-gray-dark color-blue-on-hover"
+          key={item.href}
+        >
+          {item.text}
+        </a>
+      );
+    });
+
+    return (
+      <div className="limiter flex-parent-mm flex-parent--center-cross">
```
davidtheclark: `flex-parent-mm` could just ` flex-parent`, I think.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+
+        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s">
+          {itemEls}
+        </nav>
+        <div
+          id="mbx-user-menu"
+          style={{ width: 66 }}
+          className="flex-child py6 round-full border color-blue txt-bold align-center txt-s"
+        >
+          User
```
davidtheclark: I don't think we want this placeholder.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+          </div>
+        </div>
+
+        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s">
+          {itemEls}
+        </nav>
+        <div
+          id="mbx-user-menu"
+          style={{ width: 66 }}
+          className="flex-child py6 round-full border color-blue txt-bold align-center txt-s"
```
davidtheclark: None of these are necessar except `flex-chidl`, right?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+    doc.addEventListener('click', this.closeOnDocumentClick);
+  }
+
+  componentWillUnmount() {
+    const doc = window.document;
+    doc.removeEventListener('click', this.closeOnDocumentClick);
+  }
+
+  closeOnDocumentClick = event => {
+    // Close the popup if the target is me
```
davidtheclark: me?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+    if (
+      this.state.open &&
+      event.target instanceof Node &&
```
davidtheclark: Can we remove this line? I think @jseppi only included it in user menu code to please Flow.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+    const triggerClientRect = triggerEl.getBoundingClientRect();
+    const containerClientRect = containerEl.getBoundingClientRect();
+    pointerEl.style.left =
+      Math.round(
+        triggerClientRect.left +
+          triggerClientRect.width / 2 -
+          containerClientRect.left -
+          16
+      ) + 'px';
+    containerEl.style.top =
```
davidtheclark: Since we don't have any animation, do we need this fancy post-mount positioning? Or can we use absolute positioning to get it in the right place right away?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+      top: '-16px'
+    };
+
+    const { items } = this.props;
+    const itemEls = items.map((item, i) => {
+      if (items.length === i + 1) {
+        return (
+          <li
+            key={item.href}
+            className="color-gray-dark color-blue-on-hover txt-s txt-bold block"
```
davidtheclark: Is the only different here the `mb12` class? If so, we should try to reduce the amount of repetition in this conditional.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <div className="relative">
+        <div className="limiter flex-parent flex-parent--center-cross flex-parent--space-between-main">
+          <div className="relative flex-child">
```
davidtheclark: Why is this `relative`?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+          </nav>
+        </div>
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <div className="relative">
+        <div className="limiter flex-parent flex-parent--center-cross flex-parent--space-between-main">
```
davidtheclark: Re `flex-parent--space-between-main`: We'll want to change the positioning so the logo & name is left-aligned, matching the recent move in the docs-page-shell.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+            </button>
+          </div>
+          <div className="flex-child">{this.renderLogoName()}</div>
+          <div
+            id="mbx-user-menu-mobile"
+            style={{ width: 66 }}
+            className="flex-child py6 round-full border color-blue txt-bold align-center txt-s"
+          >
+            User
+          </div>
```
davidtheclark: Some comments on this element as above.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+    const doc = window.document;
+    doc.removeEventListener('click', this.closeOnDocumentClick);
+  }
+
+  closeOnDocumentClick = event => {
+    // Close the popup if the target is me
+    if (
+      this.state.open &&
+      event.target instanceof Node &&
+      !this.node.contains(event.target)
```
davidtheclark: If `this.node` is the `<button>` node, doesn't that mean the menu will close when you click *inside* it (but maybe not on a link)?

I think we have the following requirements when the menu is open:

- If you click on a link in the menu, it closes.
- If you click the trigger, it closes.
- If you click outside the menu, it closes.
- If you click inside the menu but not on a link (e.g. you miss and hit some anodyne whitespace), it does not close.

src/components/page-header/__tests__/page-header-test-cases.js
```diff
@@ -0,0 +1,31 @@
...
+        href: '/tilesets',
+        text: 'Tilesets'
+      },
+      {
+        href: '/datasets',
+        text: 'Datasets'
+      },
+      {
+        href: '/foo',
+        text: 'Another item with a long name'
```
davidtheclark: Let's cut this because it doesn't match a real use case. These names kind of *need* to be shortish. Instead, it would be good to test the width limits of this by trying the longer menu from Account Dashboard.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,271 @@
...
+            <a href={item.href}>{item.text}</a>
+          </li>
+        );
+      }
+    });
+
+    return (
+      <div
+        id="mobile-menu-container"
+        className="absolute left shadow-darken10-bold bg-white round"
```
davidtheclark: Let's offset this from the left side just a little bit — 10 pixels, like the user menu body.

src/components/page-header/page-header.js
danswick: Ah, totally. I was thinking it should be flexible enough to deal with a parent that may or may not change height, but 1) the parent should never change height, at least given the current design and 2) the dropdown position will always be relative to the trigger, so parent height doesn't matter anyway.

src/components/page-header/page-header.js
davidtheclark: I don't think I understand that ... why not delete `relative`?

danswick:
After adding a new test case for Accounts, it looks like we may need to consider some design options in order to make things fit. 

The following screenshots are taking with page width set to 640 pixels, the cutoff for our mobile breakpoint (note that the plain component doesn't actually have a visual user menu element, so I added a silly blue box to help you visualize the spacing): 

image
_Here it is with nothing changed_ 😭 

image
_Here it is with some item name tweaks_.

- Removed "Dashboard." With the new `account` subdomain, you can just click on the big "Account" page name to go back to the dashboard.
- Changed "Access tokens" to "Tokens." 
- Changed "Statistics" to "Stats." 

For reference, the full complement of Nav items starts to look normal at around 880 pixels wide:
image


## Some ideas

Here are some things we could try to make more room: 

1. Set a higher mobile breakpoint so that the hamburger menu shows up on larger screens.
2. Have the `mapbox` part of the wordmark disappear at some point to make room for more menu items. 
3. Expand the nav to a second row at some width. 
4. Dynamically add items to a "more" dropdown like [this Microsoft documentation site](https://docs.microsoft.com/en-us/visualstudio/javascript/tutorial-nodejs-with-react-and-jsx?view=vs-2017). This will be more difficult to implement and I'm not sure it's a great idea to hide any of the Nav items for Account. 

@mapbox/accounts @angel - would you mind chiming in with your reactions, feedback, and preferences for how we might better fit these items into the Account nav?

angel:
I vote for 2. Since we cut that off on mobile anyways, I think it would be ok to cut if off a little earlier, but I like have the links still available at the larger screen size to tap easily.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,247 @@
...
+            >
+              {siteName}
+            </a>
+          </div>
+        </div>
+
+        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s mt6">
+          {itemEls}
+        </nav>
+        <div id="mbx-user-menu" style={{ width: 66 }} className="flex-child" />
```
davidtheclark: This needs `flex-child--no-shrink` to ensure it's never narrower than `66px`.

I noticed it got shrunk on one of the screenshots you posted.

danswick:
Thanks, @angel! How do you feel about dropping "Dashboard" and relying on the blue "Account" link to get users back to the main account page? If we dynamically remove "mapbox," this is as cramped as it gets: 

image


followed by: 

image

and then the most cramped before going mobile hamburger: 

image

src/components/page-header/__tests__/page-header.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+describe(testCases.basic.description, () => {
+  beforeEach(() => {
+    testCase = testCases.basic;
+  });
+
+  test('renders as expected', () => {
+    const wrapper = shallow(
+      React.createElement(testCase.component, testCase.props)
+    );
```
davidtheclark: Let's be as consistent as possible within a file about whether we create the wrapper in the `beforeEach` hook or in the `test` block.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,254 @@
...
+          </a>
+        </li>
+      );
+    });
+
+    return (
+      <div
+        id="mobile-menu-container"
+        className="absolute left shadow-darken10-bold bg-white round"
+        style={{ top: '56px', marginLeft: MOBILE_NAV_MENU_OFFSET_LEFT }}
```
davidtheclark: Where does the magic number `56` come from? Can we make it a named constant?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,254 @@
...
+                size={MOBILE_NAV_TRIGGER_WIDTH}
+              />
+            </button>
+          </div>
+          <div className="flex-child flex-child--grow ml12">
+            {this.renderLogoName()}
+          </div>
+          <div
+            id="mbx-user-menu-mobile"
+            style={{ width: 66 }}
```
davidtheclark: Let's make this magic `66` a named constant, too.

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,254 @@
...
+          {item.text}
+        </a>
+      );
+    });
+
+    return (
+      <div className="limiter flex-parent flex-parent--center-cross">
+        <div className="flex-child flex-child--no-shrink flex-parent flex-parent--center-cross">
+          <a
+            className="flex-child mb-logo wmax180-mxl wmax30-mm"
```
davidtheclark: Can `wmax30-mm` be `wmax30`, to simplify things?

src/components/page-header/page-header.js
```diff
@@ -0,0 +1,254 @@
...
+            <a
+              href="/"
+              className="flex-child color-blue color-blue-dark-on-hover ml6-mm ml0-mxl"
+            >
+              {siteName}
+            </a>
+          </div>
+        </div>
+
+        <nav className="flex-child flex-child--grow flex-parent flex-parent--center-cross flex-parent--end-main txt-bold txt-s mt6">
```
davidtheclark: What is `mt6` for? It looks to me like it throws off the vertical centering.

danswick: I think it looks kind of weird to have different sizes of text vertically centered, so I aligned their baselines instead (tho I'm sure there's a better way to do it than with hardcoded margins). Though I guess everything else in the page shell is centered vertically and my stance on the issue isn't super strong, so I can change it back. 

image

davidtheclark: I'm fine with whatever @angel and @mapbox/studio want. And if there's solid consensus we should implement the same alignment in docs-page-shell.

davidtheclark:
@danswick The document-wide click event listener that causes the closing is currently added when the whole header mounts, removed when it unmounts. What if, instead, it was added only after the menu opened, then removed after the menu closed? This would also prevent there from being an unnecessary click handler attached to the document at all times.

davidtheclark:
Another point to address would be to use a different element attribute instead of ID.

It's true that this point wouldn't be worth spending too much time on because it isn't realistic. We could have just one test cases that pushes the limits.

src/components/page-header/page-header.js
angel: I'm a fan of center alignment cause it would also be centered to the logo, button, and avatar, but can do this way if Studio feels more strongly about it.

danswick: sounds good. switching back to centered text 👍

danswick:
Ok! I think I got everything in here: 

- Text is re-centered vertically.
- The mobile menu trigger is using refs instead of IDs now.
- Document click listers are added when the menu is created and rendered when the menu is closed.
- CSS is cleaned up.
- Tests are consistified. 
- Magic numbers are `const`'d. 

Going to merge, though I'm a bit disappointed we didn't hit 50 comments.

PR #61, opened 2018-12-11
Add page-footer component

danswick:
Per discussion, pages moving away from the old page shell will have a dead-simple footer component that only includes our required legal info. App-specific footers can live above this new, legal-only footer. The component in this PR is lifted almost wholesale from the old page shell component. The only changes I made were to translate the page shell-specific classes to assembly classes and to remove the `classNames` prop in favor of the classes we've previously used on marketing pages. 

@davidtheclark for review, please 🙏

### DIFF
--- a/src/components/page-footer/page-footer.js
+++ b/src/components/page-footer/page-footer.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class PageFooter extends React.Component {
+  shouldComponentUpdate() {
+    return false;
+  }
+
+  render() {
+    return (
+      <div className="py12 py36-ml">
+        <span className="mr18">© Mapbox</span>
+        <a
+          className="link color-darken50 color-blue-on-hover mr18"
+          href="/tos/"
+        >
+          Terms
+        </a>
+        <a
+          className="link color-darken50 color-blue-on-hover mr18"
+          href="/privacy/"
+        >
+          Privacy
+        </a>
+        <a
+          className="link color-darken50 color-blue-on-hover"
+          href="/platform/security/"
+        >
+          Security
+        </a>
+      </div>
+    );
+  }
+}
+
+PageFooter.propTypes = {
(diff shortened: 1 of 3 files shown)


src/components/page-footer/page-footer.js
```diff
@@ -0,0 +1,40 @@
...
+        >
+          Security
+        </a>
+      </div>
+    );
+  }
+}
+
+PageFooter.propTypes = {
+  className: PropTypes.string
```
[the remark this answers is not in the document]
danswick: Ah, no! I accidentally left that in there from an experiment. Will 🔪

danswick:
Thanks for the feedback, @davidtheclark. I just pushed a couple of updates and have a couple of questions: 

- The footer component doesn't accept any props, which was breaking the `build-docs-data` script. I added a check to the `processProps` function to work around this -- is there a better way to go about handling prop-less components?
- Since this isn't really a footer anymore, I'm thinking we should change its name to `footer-legal-strip` as it is the dotcom page shell. What do you think?

scripts/build-docs-data.js
```diff
@@ -57,21 +57,23 @@ function getExamples(componentDirectory) {
...
+      objectBody += `${prop}: {
+        type: ${JSON.stringify(propData.type)},
+        required: ${propData.required},
+        defaultValue: ${propData.defaultValue &&
+          JSON.stringify(propData.defaultValue.value)},
+        description: ${renderedDescription}
+      },`;
+    });
+    return `{${objectBody}}`;
+  }
```
davidtheclark: This looks like the right idea to me. One change I'd suggest: to keep the function signature predictable, should we return an empty string if there are no props? We could use the early return pattern to reduce nesting.

danswick: Good call! Per chat, I returned `null` instead of an empty string to avoid a syntax error when the resulting components object is parsed in a later step.

Issue #88, opened 2019-10-11
ControlSelect component's select value does not update on change


adrianababakanian:
When selecting an `option` from the dropdown, the `select` tag's `value` does not update. 

Issue can be demonstrated with the example at: https://mapbox.github.io/mr-ui/#controlselect

Originally noticed this at the Static Images API Playground style selection (https://github.com/mapbox/playground/issues/15).

tristen:
Closing as mr-ui components will intentionally require control (i.e a new `value` must be provided by the consumer). I'm going to make the example https://mapbox.github.io/mr-ui/#controlselect more user friendly with a working implementation.

PR #89, opened 2019-10-11
Update ControlSelect so that select value is set by component state

adrianababakanian:
Fixes https://github.com/mapbox/mr-ui/issues/88


tristen:
Closing per https://github.com/mapbox/mr-ui/issues/88#issuecomment-1198609469

PR #90, opened 2019-10-21
Update badge to be green by default

HeyStenson:
This PR updates the Badge component to have a green theme by default. Reasoning: the Badge component's primary/only use is in the [dr-ui beta flag](https://github.com/mapbox/dr-ui/tree/master/src/components/beta-flag), which is often accompanied by the "new" note theme (used to discuss new/beta products), also green.

Screen Shot 2019-10-21 at 11 30 44 AM

@colleenmcginnis to review please! 🙏

### DIFF
--- a/src/components/badge/__tests__/__snapshots__/badge.test.js.snap
+++ b/src/components/badge/__tests__/__snapshots__/badge.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`badge basic renders as expected 1`] = `
 <div
-  className="txt-s txt-bold round px6 py3 inline-block cursor-default bg-orange-faint color-orange-dark "
+  className="txt-s txt-bold round px6 py3 inline-block cursor-default bg-green-faint color-green-dark"
 >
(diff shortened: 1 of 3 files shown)


HeyStenson:
> Sorry if I led you astray!

No, I for sure led myself astray! I thought that since the color for the BetaFlag component was being set here, this was the place to change it 😅 

But I should still be creating the green color scheme in this PR, and then update the BetaFlag component to use it - right? So the change I'd need to make to this PR is to make it orange by default again https://github.com/mapbox/mr-ui/pull/90/files#diff-39a2ebe0d5649f2d1cc177f850a66cb6L17? 🤔

colleenmcginnis:
>So the change I'd need to make to this PR is to make it orange by default again

Yes! I think the Accounts team also uses this component so we shouldn't mess with the defaults without consulting them. (Though aligning on colors for "beta" badges across docs/accounts/studio would be a good thing to do!)

cc @lshig

Issue #49, opened 2018-10-29
Clicking copy in CodeSnippet causes crazy scroll expansion


davidtheclark:
I noticed this behavior while checking out https://www.mapbox.com/android-docs/maps/overview/annotations/#polyline-and-polygons

![oct-28-2018 21-25-49](https://user-images.githubusercontent.com/628431/47629435-3de59200-daf8-11e8-877b-f151ffc62c4c.gif)

cc @colleenmcginnis @mapbox/frontend-platform

katydecorah:
With a copy button in a scrolled section, when you click the clipboard button the clipboard library adds a  `textarea` element to the dom, but tries to hide it through big top/left values:

![image](https://user-images.githubusercontent.com/2180540/76097210-4f189b80-5f95-11ea-82d0-ba8daff01211.png)

I think this top value is what causes the scrolled section to suddenly have a really big scroll, because when I delete the textarea from the dom, the scroll goes back to normal. I'm going to try to create a reduced test case and and create an issue in https://github.com/zenorocha/clipboard.js

katydecorah:
I think we need to hit `this.clipboard.destroy()` after copy.

katydecorah:
#105 fixed the original post, but introduced another issue: once the clipboard is destroyed it can't be copied again. The fix now is destroying the clipboard and then rebuilding it.

PR #92, opened 2019-12-05
focusTrapPaused in copyButton to work for popover use case

aparlato:
ref: https://github.com/mapbox/studio/issues/12578

ref: https://github.com/mapbox/mr-ui/issues/49

☝️ I decided not to close the above issue because while this fixes the specific issue we were having in Studio, I'm seeing that it may not fix other infinite scroll issues which may require more investigation.

Copy button was optimized for modals by trapping focus. This was causing breakage where it could lead to an infinite scroll. The copy button no longer defaults to be optimized for modals.

Also, to note: For this fix to actually fix https://github.com/mapbox/studio/issues/12578 we need to switch out [`@mapbox/react-copy-button`](https://github.com/mapbox/studio/blob/master/src/shared_components/copiable.js#L5) for `@mapbox/mr-ui/copy-button` and add the `focusTrapPaused={true}` to the `CopyButton` component.

### Question

~~The `trapFocus` prop makes the most sense to me, but it will be a breaking change since it needs to be set to `true` for modals. Should we change this prop to `pauseFocusTrap` and set that to `true` for instances that are **not** modals? That would not be breaking, but I feel like it's optimizing a lesser user case?~~

Switched to `focusTrapPaused`.

## ~~TODO~~

~~Currently working on replicating the original issue in test cases app to demonstrate this fix works.~~

Per conversation, we're going to see if this works and not drive ourselves insane to get a replicating test case here. Logically this makes sense.

cc @elifitch

### DIFF
--- a/src/components/copy-button/__tests__/copy-button.test.js
+++ b/src/components/copy-button/__tests__/copy-button.test.js
@@ -3,15 +3,27 @@ import { shallow } from 'enzyme';
 import delay from 'delay';
 import { testCases } from './copy-button-test-cases';
 import getWindow from '../../utils/get-window';
+import Clipboard from 'clipboard/dist/clipboard.min.js';
 
 const mockUserAgent =
   'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36';
 
-jest.mock('clipboard/dist/clipboard.min.js', () => {
-  class Clipboard {}
-  Clipboard.isSupported = () => true;
-  return Clipboard;
-});
+jest.mock('clipboard/dist/clipboard.min.js');
+
+// jest.mock('clipboard/dist/clipboard.min.js', () => {
+//   class Clipboard {}
+//   Clipboard.isSupported = () => true;
+//   return Clipboard;
+// });
+
+// const FakeClipboard = jest.fn(() => {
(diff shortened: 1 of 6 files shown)


aparlato:
@elifitch I made an edit to the original comment about not closing the original `mr-ui` issue. I'm seeing odd scrolling behavior still present with a `CopyButton` in a `CodeSnippet` that doesn't totally replicate the original issue, but isn't changed by this fix. Due to difficulty replicating the original issue in a test environment and the fact that this generally seems like a good fix and solves our problem in Studio, I still think we should move forward. I suspect there might be wider spanning issues/patterns in `mr-ui` around focus that are causing various scrolling issues.

src/components/copy-button/__tests__/copy-button.test.js
```diff
@@ -3,15 +3,27 @@ import { shallow } from 'enzyme';
...
   'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36';
 
-jest.mock('clipboard/dist/clipboard.min.js', () => {
-  class Clipboard {}
-  Clipboard.isSupported = () => true;
-  return Clipboard;
-});
+jest.mock('clipboard/dist/clipboard.min.js');
+
+// jest.mock('clipboard/dist/clipboard.min.js', () => {
```
elifitch: 🔪

PR #104, opened 2020-03-03
Add aria to all button components, examples, and tests

lshig:
I've been monitoring our Lighthouse Accessibility score and see some opportunities to improve our buttons. For example:

> When a button doesn't have an accessible name, screen readers announce it as "button", making it unusable for users who rely on screen readers. [Learn more.](https://web.dev/button-name/)"

The account dashboard mostly uses `CopyButton` and `Copiable`, but this PR looks to change all components, examples, and tests that use `<button>` or our custom `Button` component to encourage ARIA for now. For `<button>` the `aria-label` defaults to the `label` or `id` provided to the button (whichever is a string); if a `Tooltip` is wrapping the `<button>`, then the `aria-label` will match the text in the `Tooltip`. For `Button`, users will need to use `passthroughProps` and determine if they want to use `aria`; though all our examples and test cases will encourage `aria` for our own accessibility purposes.

### DIFF
--- a/src/components/button/__tests__/button-test-cases.js
+++ b/src/components/button/__tests__/button-test-cases.js
@@ -12,7 +12,7 @@ testCases.primary = {
     children: 'Primary',
     onClick: safeSpy(),
     passthroughProps: {
-      'aria-label': 'Primary'
+      'aria-label': 'Example primary button'
     }
   }
 };
@@ -25,7 +25,7 @@ testCases.secondary = {
     variant: 'secondary',
     onClick: safeSpy(),
     passthroughProps: {
-      'aria-label': 'Secondary'
+      'aria-label': 'Example secondary button'
     }
   }
 };
@@ -38,7 +38,7 @@ testCases.discouraging = {
     variant: 'discouraging',
     onClick: safeSpy(),
     passthroughProps: {
-      'aria-label': 'Discouraging'
+      'aria-label': 'Example discouraging button'
     }
   }
 };
@@ -51,7 +51,7 @@ testCases.destructive = {
     variant: 'destructive',
     onClick: safeSpy(),
     passthroughProps: {
-      'aria-label': 'Destructive'
+      'aria-label': 'Example destructive button'
     }
   }
 };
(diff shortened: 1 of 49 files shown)


src/components/form-submit/form-submit.js
```diff
@@ -32,7 +32,7 @@ export default class FormSubmit extends React.PureComponent {
     };
 
     return (
-      <button type="button" className={themeButton} {...inputProps}>
+      <button aria-label={label} type="button" className={themeButton} {...inputProps}>
```
danswick: Should we make `label` a required propType to ensure the `aria-label` is always present?

src/docs/components/component-example.js
```diff
@@ -53,6 +53,7 @@ function ToggleCodeButton(props) {
   const text = props.codeIsVisible ? 'Hide code' : 'Show code';
   return (
     <button
+      aria-label={text}
```
danswick: Do you know if screen reading software has a preference between `aria-label` and `aria-labelledby`? For cases like this one, it seems like we can reliably point to the element that contains the text, but I'm not sure if there is anything to be gained by the additional complexity of adding an ID to the span that contains the `{text}`.

ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

src/components/button/__tests__/button-test-cases.js
```diff
@@ -50,7 +62,10 @@ testCases.appPrimary = {
   props: {
     children: 'AppPrimary',
     variant: 'appPrimary',
-    onClick: safeSpy()
+    onClick: safeSpy(),
+    passthroughProps: {
+      'aria-label': 'AppPrimary'
```
danswick: I know these are just test cases, but maybe we could use an example here that demonstrates what a useful label would look like? A label like this would probably be of limited utility to someone using a screen reader.

src/components/form-submit/form-submit.js
lshig: `FormSubmit` has `label: 'Submit'` in the `defaultProps` so we should be ok to fallback on it for this component and not require a prop `label` here.

src/docs/components/component-example.js
lshig: `aria-label` will take precedence over the `{text}` or children elements within `<button>` and `aria-labelledby` will take precedence over `aria-label` if they both exist on the same element.  Instead of applying `aria-labelledby` to `<button>`, it looks like we would want to utilize `aria-labelledby` for other types of form `<input>`.

Issue #85, opened 2019-09-11
Click events inside of tooltips inside of modals getting eaten


chrisbrainerd:
Minimal example: https://codesandbox.io/s/exciting-bassi-8zye4?fontsize=14

Originally thought it was only anchor tags but I tried it with a button's onClick and that doesn't work either, both get eaten.

dereklieu:
This is pretty old so maybe no longer relevant to you @chrisbrainerd, but popping in to say that we had this problem in Studio, and added a `focusTrapPaused` prop to the [modal](https://mapbox.github.io/mr-ui/#modal) component - when set to true, the inputs in that example will work.

tristen:
Haven't confirmed but I'm going to close as stale. The internal ingredients have been swapped out.

PR #111, opened 2020-09-03
Stop event propagation outside of modals by default

elifitch:
This allows for simple, intuitive Modal nesting without having to worry about which element the modal is attached to, or worrying about differences between the virtual dom and browser dom.

We do this by stopping event bubbling outside of modals by default, and adds a prop to enable event bubbling outside of a modal should you need that functionality. Bubbling is disabled by default because that was the previous behavior under React 15, so this isn't a breaking change, and it's IMO more intuitive to not have events propagate outside of what is effectively a "UI island".

I made a barebones event trapping component, which stops propagation on all keyboard, focus, form, touch and mouse events from https://reactjs.org/docs/events.html. This <EventTrap> component wraps <AriaModal> and keeps events contained.

This PR also targets a new React 16 feature branch were we can collect React 16 changes to one day release them all as a single major version bump.

@dereklieu for review


src/components/modal/__tests__/modal-test-cases.js
```diff
@@ -34,6 +39,90 @@ class ModalWrapper extends React.Component {
...
+        {modal}
+      </div>
+    );
+  }
+}
+
+testCases.nested = {
+  description: 'nested modals',
+  element: <OuterModal />
+};
```
dereklieu, line 124: I'm not familiar with how test cases here work. Where is this test case being used? I'm not seeing a new snapshot, did I miss it?

dereklieu (commented):
> This allows for simple, intuitive Modal nesting without having to worry about which element the modal is attached to

I've been wondering, are nested modals an anti-pattern that this change will further enable? Should we go the other way and stop building nested modals? Judging from our issue queue, we've already removed the one instance where we were using this pattern in https://github.com/mapbox/studio/pull/13446.

Reasons why we might want to keep this behavior in place:

- Focus trapping gets weird. Issues like this one, and https://github.com/mapbox/mr-ui/issues/85, show to me that popovers and modals, as implemented in this repo, are not optimal for tandem or multiple use. 
- Portals will find their way into the Studio codebase, and we want to be aware and consistent on their behavior throughout our dependency tree.

Also: have you tried this in Studio and confirmed that it fixes our problem with https://github.com/mapbox/studio/pull/13446 reverted?

elifitch:
> are nested modals an anti-pattern that this change will further enable

Maybe, but I see them as nearly inevitable. If we want to have freely sharable modals for stuff like confirmations or error states, and if we ever put UI flows inside of a modal, we'll inevitably run into situations where we are nesting modals. If we want to avoid nested modals, but still have basic confirmation/error state modals, we'd have to move all significant UI flows outside of modals. That would be a pretty big undertaking in Studio and probably other apps too.

> Focus trapping gets weird.

elifitch:
When using this in studio we run into the same problem we got before, where the buttons on the inner modal don't respond. That, and the difference in behavior between the dataset list page and the dataset editor seems to indicate that there's something in how we're composing this component in studio that's wonky. I'll need to dig some more.

elifitch:
Studio WIP pushed to https://github.com/mapbox/studio/pull/13464. When NPM linking mr ui, remember that they both need to be on the same node version, and because mr-ui builds to `pkg/` you'll need to run `npm link` from within that directory, not root.

dereklieu:
> If we want to have freely sharable modals for stuff like confirmations or error states, and if we ever put UI flows inside of a modal, we'll inevitably run into situations where we are nesting modals.

Have you checked out [the tileset migration modal](https://github.com/mapbox/studio/blob/main/src/styles_components/migrate_style/migrate_modal_content.js)? This is an example of a stateful modal that renders confirmation states without nesting additional modals. 

I agree that freely using modals anywhere we like is convenient, but my point with focus trapping being weird is that this has been a source of bugs for us, and I think it will continue to be after this PR.

Curious if anyone else at @mapbox/map-design-studio has strong opinions here.

elifitch:
Hm I see this as kind of distinct from focus trapping, I see it as building a moat around events in each modal. To make sure that we don't accidentally trigger events in a parent, since a Modal should be a self-contained UI piece. What kinds of problems do you foresee with this barrier around event bubbling from modal to modal?

samanpwbb:
Drive by thoughts on this topic: 
- There's some inherent complexity to focus handling with multiple modals that we could deal with by using a system where we only ever render 1 modal at a time, but modals could be queued in the background. This way, parent modals could still call children modals, and then re-open once user is done with the child modal.
- In general, I do think we over-use modals in studio for confirmation and notification and – there are some nice UI patterns for inline confirmation and then 'toast' style patterns for success notifications we could be using.

dereklieu (approved):
👍

src/components/modal/__tests__/modal-test-cases.js
elifitch: This test case is in the test cases app. You're supposed to be able to use these test cases for both unit tests and for our janky storybook esque implementation.

PR #113, opened 2020-09-08
Only render "More +" in tab list when needed

adrianababakanian:
Currently, the `moreButton` of the tab list component will always render regardless of whether there are actually any additional items beyond those already visibly enumerated. This PR updates the behavior to only render the button when `truncateBy < items.length`.

### DIFF
--- a/src/components/tab-list/__tests__/tab-list.test.js
+++ b/src/components/tab-list/__tests__/tab-list.test.js
@@ -50,7 +50,7 @@ describe('TabList', () => {
     test('activeItem prop updates active item', () => {
       wrapper.setProps({ activeItem: 'one' });
       wrapper.update();
-      expect(wrapper.find('.border--blue.txt-bold').length).toBe(1);
+      expect(wrapper.find({ 'data-test': 'one' }).length).toBe(1);
     });
   });
   describe(testCases.truncateAll.description, () => {
@@ -68,7 +68,7 @@ describe('TabList', () => {
     test('activeItem prop updates active item', () => {
       wrapper.setProps({ activeItem: 'one' });
       wrapper.update();
-      expect(wrapper.find('.border--blue.txt-bold').length).toBe(1);
+      expect(wrapper.find({ 'data-test': 'one' }).length).toBe(1);
     });
   });
   describe(testCases.moreButtonOnlyIfNeeded.description, () => {
@@ -86,7 +86,7 @@ describe('TabList', () => {
     test('activeItem prop updates active item', () => {
       wrapper.setProps({ activeItem: 'one' });
       wrapper.update();
-      expect(wrapper.find('.border--blue.txt-bold').length).toBe(1);
+      expect(wrapper.find({ 'data-test': 'one' }).length).toBe(1);
     });
   });
   describe(testCases.labelNode.description, () => {
@@ -114,7 +114,7 @@ describe('TabList', () => {
     test('activeItem prop updates active item', () => {
       wrapper.setProps({ activeItem: 'one' });
       wrapper.update();
-      expect(wrapper.find('.border--blue.txt-bold').length).toBe(1);
+      expect(wrapper.find({ 'data-test': 'one' }).length).toBe(1);
     });
   });
 });
--- a/src/components/tab-list/__tests__/tab-list.test.js
+++ b/src/components/tab-list/__tests__/tab-list.test.js
@@ -72,6 +72,25 @@ describe('TabList', () => {
       wrapper.update();
     });
   });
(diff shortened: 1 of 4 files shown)


elifitch:
@adrianababakanian Given our current test cases, it's hard for me to get myself in a situation where I can see the problem this PR fixes. Could you make another test case that shows off this improved behavior, and will help prevent regressions?

adrianababakanian:
Sure thing @elifitch , great suggestion -- I'll add that test to the PR and then follow up. Thank you!

adrianababakanian:
Hi @elifitch , just added a test + snapshot! Hopefully the GIFs in this comment help better demonstrate the change that this PR introduces as well -- let me know if anything is unclear and I would be more than happy to work to improve it. The only relevant test case in the GIFs below is the one with the description "only show the more + button if there are more items not currently visible."

## Before
_The "More + " button is extraneously added to the `TabList` with only two items on narrower screens._

tablist-before(1)


## After
_The "More + " button is not added when the screen narrows, as expected._

tablist-after

src/components/tab-list/__tests__/tab-list.test.js
```diff
@@ -72,6 +72,25 @@ describe('TabList', () => {
...
+      );
+    });
+
+    test('renders', () => {
+      expect(
+        toJson(shallow(React.createElement(testCase.component, testCase.props)))
+      ).toMatchSnapshot();
+    });
+
+    test('activeItem prop updates active item', () => {
```
elifitch, line 89: This test needs an assertion, looks like you missed an `expect` in here somewhere

src/components/tab-list/__tests__/tab-list.test.js
```diff
@@ -72,6 +72,25 @@ describe('TabList', () => {
...
+    beforeEach(() => {
+      testCase = testCases.moreButtonOnlyIfNeeded;
+      wrapper = shallow(
+        React.createElement(testCase.component, testCase.props)
+      );
+    });
+
+    test('renders', () => {
+      expect(
+        toJson(shallow(React.createElement(testCase.component, testCase.props)))
```
elifitch, line 85: Could tidy this up with `toJson(wrapper)` since wrapper is already defined as the result of `React.createElement(testCase.component, testCase.props)`

elifitch (approved):
Thanks for gif and test case! Much clearer now what's going on. Had a few little pieces of feedback about the test, but once those are resolved, this should be good to go.

src/components/tab-list/__tests__/tab-list.test.js
adrianababakanian: Interesting, I actually just copied this particular test verbatim from how it is included in each of the other `describe` blocks in this test suite. I'll push a commit to add an `expect` both here and in the other `describe` blocks!

src/components/tab-list/__tests__/tab-list.test.js
adrianababakanian: Good catch, I'll make that update here and in the other test cases.

adrianababakanian:
Thanks for the review @elifitch ! I pushed 37e69952510065eb64ab104eff61db81871db2c0 which addresses your feedback on the test cases. Since I had copy-pasted these testing particulars from existing tests in this suite, I went ahead and applied the updates to the other relevant tests as well. As such, I would appreciate your feedback one more time before merging.

In particular, I chose to use `.border--blue.txt-bold` explicitly as the selector for `wrapper.find`, since this is the default value for the `themeActiveItem` prop ([ref](https://github.com/mapbox/mr-ui/blob/a6ff96a028dca71cd8a66241fc6426cdb0f37d70/src/components/tab-list/tab-list.js#L48)) which is subsequently used as the `activeClasses` for the `activeItem`:

https://github.com/mapbox/mr-ui/blob/a6ff96a028dca71cd8a66241fc6426cdb0f37d70/src/components/tab-list/tab-list.js#L87-L89

Just wanted to get your opinion on whether hardcoding this value looks alright here (since each of the test cases use the default value for the `themeActiveItem` prop and don't set it explicitly), or whether we should get the prop's value from the wrapper dynamically in the test to use as the selector.

elifitch:
@adrianababakanian rather than using a CSS selector for this, which would change if we ever updated the presentation of this component, we usually add a `data-test` attribute ala https://github.com/mapbox/mr-ui/blob/fb9f4911c805265b7d6df54fd60e4023169e3f95/src/components/modal/modal-actions.js#L18. We should probably do that here as well. Other than this, my approval stands. Go ahead and make this little update and then merge!

PR #114, opened 2020-09-23
Set `devServerHistoryFallback: true` to use HTML5 History for client-side routing

katydecorah:
This PR enables [devServerHistoryFallback](https://github.com/mapbox/underreact#devserverhistoryfallback) on the test cases app. This solves the problem where when you refresh a test case app component page, the app fails to fetch the route.

## How to test

1. Run `npm start`
2. Click on any component link.
3. Refresh the page, the component test case app should refresh and display properly.


elifitch:
Works great @katydecorah! Can't believe this was a simple one line fix!

elifitch (approved):

PR #115, opened 2020-10-05
Add `inline` prop to IconText component

adrianababakanian:
This PR adds an optional `inline` prop to the `IconText` component, with a default value of `false`. When set to `true`, the outer `<span>` wrapping the component will have a class of `'flex-parent-inline'` rather than the default `'flex-parent'`. This change will allow the `IconText` component to be horizontally centered as expected when used as the child of an element with class `'align-center'`.

Note that one can't get around this (though please do correct me if I'm wrong!) by adding a `className` prop with a value set to `'mx-auto'` since the outer `<span>` still would have `display: inline`.


katydecorah (approved):
Looks great, @adrianababakanian! I think this addition makes a lot of sense.

PR #140, opened 2021-10-04
Drop non existing tooltip props and pass popoverProps

### LINKED ISSUE #119: input error tooltips reference props that don't exist
The Popover component doesn't acknowledge these props https://github.com/mapbox/mr-ui/blob/a6cb54c320a48b8080d5e29ffb68ee437a28d1b6/src/components/control-text/control-text.js#L170-L176

So the tooltip is not styled

Screen Shot 2021-04-04 at 4 48 56 PM

tristen:
| Before | After |
| --- | --- |
| before |  Screen Shot 2021-10-04 at 7 34 52 PM |

Closes #119


adrianababakanian (approved):
🍏

PR #146, opened 2021-12-13
<PopoverTrigger> Gate showBecauseClick if element is inside the popover content

mbullington:
# Description

This fixes a race condition where if a button inside the popover content closes the popover on its `onClick` handler (without a microtask, etc), then it doesn't work since the popover reshows itself.


tristen (approved):
LGTM

PR #147, opened 2021-12-13
Drop xtend

### LINKED ISSUE #136: Drop xtend
Branching off this: https://github.com/mapbox/studio/issues/14905

Unless anyone using `mr-ui` (and plans on updating) needs IE11 support, I feel we should drop `xtend` since [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) is very well supported in modern browsers and iOS 9+.

mbullington:
# Description

Closes #136

These drop `xtend` in favor of spread syntax (`{ ...a }`), which we already use in `mr-ui` so this shouldn't change our browser support.

cc @chrisbrainerd


chrisbrainerd (approved):
Lgtm, but did you `npm uninstall` or just remove from the package.json? I'm still seeing it in package-lock.json but unsure if that's from other deps that use it.

mbullington:
Oh good call, I had just removed it but it's also pulled in by lots of dependencies.

PR #148, opened 2021-12-13
Use ResizeObserver if available in <PopoverPositioner>

### LINKED ISSUE #130: Investigate observing size updates of Popover child
# Problem

Right now, if the child of a `Popover` changes it's width/height, the calculations for the Popover's alignment stay the same. There is `Popover.repositionPopovers` which is helpful for scenarios you know this may occur, but tracking these scenarios individually may be bug-prone.

# Potential solution

A pattern I found that might be more generally useful is observing the child of a `Popover` with [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), so when the size updates the popover can reposition.

Maybe there could be an optional `Popover` field such as `observeSize`?

mbullington:
Closes #130 

This ticket adds an `observeSize` prop to `Popover` and `PopoverPositioner` (true by default) that observes the popover content using `ResizeObserver` and recalculates size dynamically.

This is a progressive enhancement so if `ResizeObserver` isn't available in a browser this is not used. It has extremely high browser support though, minus IE11. https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

---

This lets us also remove Studio logic that it really shouldn't need to worry about:

https://github.com/mapbox/studio/blob/main/src/shared_components/resize_observer_component.tsx
https://github.com/mapbox/studio/blob/main/src/notification_components/notification_display.js


tristen (approved):
Looks great!

PR #152, opened 2022-01-06
Reinforced password policy for reset password

lawsberryPi:
Resolve Pentest Vulnerability-2021: Weak Password Policy, Reinforced password rules so that the password must have:
one uppercase letter 
one lowercase letter
one digit
one special character 
at least 8 chars long

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "@mapbox/mr-ui",
-  "version": "2.0.0-beta.1",
+  "version": "2.0.0-beta.2",
   "description": "UI components for Mapbox projects",
   "main": "index.js",
   "scripts": {
@@ -47,6 +47,7 @@
     "focus-trap": "^3.0.0",
     "hoverintent": "^2.0.0",
     "os-key": "^1.0.0",
+    "password-validator": "^5.2.1",
     "prefix": "^1.0.0",
     "react-aria-modal": "^4.0.0",
     "react-displace": "^2.3.0",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "focus-trap": "^3.0.0",
     "hoverintent": "^2.0.0",
     "os-key": "^1.0.0",
-    "password-validator": "^5.2.1",
     "prefix": "^1.0.0",
     "react-aria-modal": "^4.0.0",
     "react-displace": "^2.3.0",
--- a/src/components/form/validators/__tests__/validate-password.test.js
+++ b/src/components/form/validators/__tests__/validate-password.test.js
@@ -1,4 +1,6 @@
 import validatePassword from '../validate-password';
+import { shallow } from 'enzyme';
+import React from 'react';
 
 test('accepts empty value', () => {
   expect(validatePassword('')).toBe('');
@@ -8,26 +10,55 @@ test('valid returns ""', () => {
   expect(validatePassword('Mapbox123~')).toBe('');
 });
 
-test('invalid missing symbol returns error', () => {
-  expect(validatePassword('Mapbox123')).toContain(
-    'The password must have a punctuation'
+test('Invalid password is too short returns error message', () => {
+  const result = shallow(validatePassword('Mbx12~')).contains(
+    <li className="list-group-item">
+      Password must be at least eight characters long.
+    </li>
   );
+  expect(result).toBeTruthy();
 });
 
-test('invalid too short returns error', () => {
-  expect(validatePassword('Mbx12~')).toContain(
-    'The password must be at least eight characters long'
+test('Invalid password missing number returns error message', () => {
+  const result = shallow(validatePassword('MAPbox@~')).contains(
+    <li className="list-group-item">Password must have a number.</li>
   );
+  expect(result).toBeTruthy();
 });
 
-test('invalid missing uppercase returns error', () => {
-  expect(validatePassword('mapbox123~')).toContain(
-    'The password must have a uppercase letter'
+test('Invalid password missing special characters returns error message', () => {
+  let element = shallow(validatePassword('Mapbox123')).contains(
+    <li className="list-group-item">
+      Password must have a special character like ".!@#$%^&*()_+-=,".
+    </li>
   );
+  expect(element).toBeTruthy();
 });
 
-test('invalid missing lowercase letter returns correct error message', () => {
-  expect(validatePassword('MAPBOX123~')).toContain(
-    'The password must have a lowercase letter'
+test('Invalid password missing uppercase returns error message', () => {
+  const result = shallow(validatePassword('mapbox123~')).contains(
+    <li className="list-group-item">Password must have an uppercase letter.</li>
   );
+  expect(result).toBeTruthy();
+});
+
+test('Invalid password missing lowercase letter returns correct error message', () => {
+  const result = shallow(validatePassword('MAPBOX123~')).contains(
+    <li className="list-group-item">Password must have a lowercase letter.</li>
+  );
+  expect(result).toBeTruthy();
+});
+
+test('Invalid password missing multiple password rules returns correct error messages', () => {
+  const result = shallow(validatePassword('~')).contains(
+    <li className="list-group-item">Password must have a lowercase letter.</li>,
+    <li className="list-group-item">Password must have a number.</li>,
+    <li className="list-group-item">
+      Password must have an uppercase letter.
+    </li>,
+    <li className="list-group-item">
+      Password must be at least eight characters long.
+    </li>
+  );
+  expect(result).toBeTruthy();
 });
--- a/src/components/form/validators/__tests__/validate-password.test.js
+++ b/src/components/form/validators/__tests__/validate-password.test.js
@@ -9,25 +9,25 @@ test('valid returns ""', () => {
 });
 
 test('invalid missing symbol returns error', () => {
-  expect(validatePassword('Mapbox123')).toBe(
-    'The string should have a minimum of 1 symbol'
+  expect(validatePassword('Mapbox123')).toContain(
+    'The password must have a punctuation'
   );
 });
 
 test('invalid too short returns error', () => {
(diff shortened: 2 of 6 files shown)


hibsy (approved):

vatolvan:
This is only UI change, right? 

Are you gonna add the corresponding change to backend side as well?

src/components/form/validators/__tests__/validate-password.test.js
```diff
@@ -5,11 +5,29 @@ test('accepts empty value', () => {
 });
 
 test('valid returns ""', () => {
-  expect(validatePassword('12345678')).toBe('');
+  expect(validatePassword('Mapbox123~')).toBe('');
+});
+
+test('invalid missing symbol returns error', () => {
+  expect(validatePassword('Mapbox123')).toBe(
+    'The string should have a minimum of 1 symbol'
```
vatolvan, line 13: This text is not very user friendly, specifically "the string" -> "the password". 

Also, what does a "symbol" mean?  I tried to think how I would resolve if I got this error myself and really have no idea 😂

package.json
```diff
@@ -47,6 +47,7 @@
     "focus-trap": "^3.0.0",
     "hoverintent": "^2.0.0",
     "os-key": "^1.0.0",
+    "password-validator": "^5.2.1",
```
vatolvan, line 50: What do you think about using regex instead of adding a 3rd party librar? The rules seem to be quite simple so the regex should not be too complicated to to

src/components/form/validators/__tests__/validate-password.test.js
Miroko: Usually these are called special characters, we could maybe even show an example for the user what a special character means. Probably any other character that is not a number or letter is considered as special character

vatolvan: Right!

I agree that it should show the accepted symbols / special characters if this check fails. Otherwise its quite hard to fix

lawsberryPi: that make sense! I will change the message

package.json
lawsberryPi: yeah, i can do that too

yseo93 (approved):

src/components/form/validators/__tests__/validate-password.test.js
```diff
@@ -5,11 +5,29 @@ test('accepts empty value', () => {
...
+  );
+});
+
+test('invalid missing uppercase returns error', () => {
+  expect(validatePassword('mapbox123~')).toContain(
+    'The password must have a uppercase letter'
+  );
+});
+
+test('invalid missing lowercase letter returns correct error message', () => {
```
vatolvan, line 29: Can you add one test for testing a case when each of the condition fails

src/components/form/validators/validate-password.js
```diff
@@ -3,9 +3,19 @@ import isEmptyControlValue from './is-empty-control-value';
 export default function validatePassword(value) {
   if (isEmptyControlValue(value)) return '';
 
-  if (value.length < 8) {
-    return 'Please enter a password with 8 or more characters';
-  }
+  if (value.length < 8)
+    return 'The password must be at least eight characters long';
```
vliman, line 7: I'm not sure it's really needed to start the message from `The` article? Maybe just `Password must be at least eight characters long.` would sound better? Also added dot to the end to be consistent with the rest of the messages.

src/components/form/validators/validate-password.js
```diff
@@ -3,9 +3,19 @@ import isEmptyControlValue from './is-empty-control-value';
...
 
-  return '';
+  let validationResult = '';
+
+  if (!/[a-z]/.test(value))
+    validationResult += 'The password must have a lowercase letter. \r\n';
+  if (!/[A-Z]/.test(value))
+    validationResult += 'The password must have a uppercase letter. \r\n';
```
vliman, line 14: `a uppercase` -> `an uppercase`, no?

src/components/form/validators/validate-password.js
```diff
@@ -3,9 +3,19 @@ import isEmptyControlValue from './is-empty-control-value';
...
+  if (!/\d/.test(value))
+    validationResult += 'The password must have a number. \r\n';
+  if (!/\W/.test(value))
+    validationResult += 'The password must have a punctuation. \r\n';
```
vliman, line 18: Wondering whether `punctuation` would be clear for end users? Maybe we should say `special characters` and list all of the acceptable characters in parentheses? In some places I also saw words `metacharacters` or `non word characters` used to describe these type of symbols, but not sure they better than `special characters`.

src/components/form/validators/validate-password.js
```diff
@@ -3,9 +3,19 @@ import isEmptyControlValue from './is-empty-control-value';
...
+
+  return validationResult;
```
vliman, line 20: Thinking about that approach and maybe we can do it in a bit different way without strings concatenation? And I do think we need to immediately return if password is less than 8 characters. As a user I'd rather prefer to see all the missed requirements straightaway.
Something like this (taking into account most of my previous comments except the one regarding punctuation)
```
export default function validatePassword(value) {
  if (isEmptyControlValue(value)) return '';

  const validations = [];

  if (value.length < 8) {
    validations.push('Password must be at least eight characters long.');
  }

  if (!/[a-z]/.test(value)) {
    validations.push('Password must have a lowercase letter.)';
  }

  if (!/[A-Z]/.test(value)) {
    validations.push('Password must have an uppercase letter.)';
  }

  if (!/\d/.test(value)) {
    validations.push('The password must have a number.)';
  }

  if (!/\W/.test(value)) {
    validations.push('The password must have a punctuation.)';
  }

  return validations.join('\r\n');
}

```

src/components/form/validators/validate-password.js
vliman: We also need to think about what special characters we allow. `\W` is equivalent to `[^a-zA-Z0-9_]` - it means if we use `\W` regex, `_` (underscore symbols) will be excluded from check. Maybe it's better to just list all of the acceptable symbols explicitly in regex? Not sure about that though.

src/components/form/validators/__tests__/validate-password.test.js
vliman, line 29: NIT: I noticed that message ends with `returns correct error message`, but the rest of the messages above end with just `returns error`. I think it make sense to make it consistent everywhere.

src/components/form/validators/validate-password.js
lawsberryPi: how about we allows users to use these special characters ? https://docs.oracle.com/cd/E11223_01/doc.910/e11197/app_special_char.htm#MCMAD416

src/components/form/validators/validate-password.js
lawsberryPi: sure, that would work

src/components/form/validators/validate-password.js
lawsberryPi: Good catch, changing now

src/components/form/validators/validate-password.js
lawsberryPi: actually I did some research, I think \W should work, I will just modify the error message just incase user don't know what special characters are

vliman: Sounds good, thanks!

mbullington (changes requested):
Approved on my end after the other comments are resolved. Just make sure to update the package lock to `2.0.0-beta.2` as well before publishing, otherwise it'll just overwrite beta 1.

Thanks for the tag!

src/components/form/validators/validate-password.js
```diff
@@ -1,11 +1,40 @@
+import React from 'react';
```
vatolvan, line 1: While technically allowed, you are now outputting JSX from pure javascript function. 

You should use the pure javascript functions to output javascript objects / data types and wrap them in JSX in your actual JSX render components / functions.

vatolvan: So instead of wrapping the list to JSX here in this component, this component should do only whats necessary. Perhaps only generate the array of validation error messages which can then be used as fit from outside this function

vatolvan: Checked the file and seems we are using the `.js` extension for everything :(

Perhaps this is ok then...

vatolvan: I would still perhaps just return a list here and use that list / array in the caller component/function

lawsberryPi: the caller is here, I feel this block of code has much more people use it, so I don't want to change it... I feel it more responsibility to change this part of code         [typeof validationError === 'string'](https://github.com/mapbox/mr-ui/blob/c62882528c3aca00c06526dbcb444eb519b72bf6/src/components/control-wrapper/control-wrapper.js#L31)

vatolvan: Discussed this in Slack with @lawsberryPi  and its quite a big change to refactor this to return lists in the scope of this PR so this is ok 👍

vliman: Can I get more clarity here, please? Previously it was a function returning string, now we are making it returning either string or markup. Can I also get examples where that function is being used?

vliman:
@lawsberryPi I think you also need to update [README](https://github.com/mapbox/mr-ui/blob/main/src/components/form/validators/README.md#validate-password) to reflect latest changes.

src/components/form/validators/validate-password.js
vliman: ok, found this [example](https://github.com/mapbox/mr-ui/blob/main/src/components/form/validators/examples/util-validators-basic.js#L94-L97). If I understood it correctly ControlText component has a property called [validationError](https://github.com/mapbox/mr-ui/blob/main/src/components/control-text/control-text.js#L57) which is of type `PropTypes.node`. So we should be fine here.

PR #161, opened 2022-06-29
Introduce TypeScript and convert two components

### LINKED ISSUE #144: TypeScript typings
# Description

I know we have discussed this before but didn't see it in a ticket so wanted to get it "on the books."

tristen:
This PR is large which is mostly due to the package-lock changes. There are - however - a number of files that have been affected as part of this work.

- Rewrites `src/components/icon` as a functional component and uses [`accessible-icon` from Radix](https://www.radix-ui.com/docs/primitives/utilities/accessible-icon) as an internal primitive.
- Rewrites `src/components/control-wrapper` as a functional component.
- Introduces TypeScript to the source code and converts:
  - `src/docs/*`
  - `src/components/utils/*`
  - `src/components/icon/*`
  - `src/components/control-wrapper/*`
- Adds React Testing Library to newly ported components

---

**Note** I am not merging this branch into `main` but into an `mr-ui@3.0.0` feature branch.

Closes #144


badiuoanaalexandra:
Do you mind if I add .nvmrc file to your branch? Seems that it needs npm version 16 for it to work?

src/components/control-wrapper/control-wrapper.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <div className={wrapperClasses}>
+      {children}
+      <div role="alert">{errorEl}</div>
+    </div>
+  );
+}
+
+ControlWrapper.propTypes = {
```
badiuoanaalexandra, line 46: Do we still need prop-types if we use typescript?

badiuoanaalexandra: I am still a bit of a newbie to Typescript (been using it for like half a year), I haven't used Typescript and prop-types in the same time and haven't yet used typescript in an own build library before. Now that I think about it, I guess it is still very useful for people who just use plain JS instead of TS, right? After reading a bit it looks like there might be some packages out there that generate prop types from TS we could later on investigate so that we don't need to write these twice. Also did not know that you use it to generate documentation, just realised that this is a public repo, so anyone can use it :) just ignore my comment :D

badiuoanaalexandra (approved):

PR #167, opened 2022-07-05
Rewrite control-switch

### LINKED ISSUE #164: Rewrite control-switch
To:

- TypeScript
- Functional component
- Radix innerworkings. Here's the API Reference: https://radix-website-git-release.modulz-deploys.com/docs/primitives/components/switch#root

### Caveats

- The current interface today allows you to pass extra props in an input element https://github.com/mapbox/mr-ui/blob/6a9074563941dec6ac7a0c4728f777bd34a8adf9/src/components/control-switch/control-switch.js#L59 which we, while Radix provides one when used within a `form` to ensure events propagate correctly, we **don't** have access to and is considered an implementation detail. I think it should be fine to narrow the prop options to switch, it just needs to be captured in documentation.
- The `input` element that is inserted using Radix is ordered below the trigger element which breaks the adjacent sibling selector Assembly uses for styling the active state. The correct way to do this is to use the selector `[data-state="checked"]`, so there's some upstream work in Assembly to handle this.

tristen:
This PR rewrites `src/components/control-switch` as a functional component.and uses [`swtich` from Radix](https://radix-website-git-release.modulz-deploys.com/docs/primitives/components/switch). 

🚨 **Breaking changes**:

  - Dropped `optional` prop: This only had the behavior of appending the words "(optional)" next to the label prop which I believe is entirely unnecessary as you can do the same thing with `label`.
  - I am ignoring the following props [previously passed](https://github.com/mapbox/mr-ui/blob/6a9074563941dec6ac7a0c4728f777bd34a8adf9/src/components/control-switch/control-switch.js#L17-L19) directly to the input element from the form component: `initialValue` and `validator`. These are undocumented, confusingly added to an HTML element, We don't have direct access to an input element from Radix and we should probably rethink the approach overall of forms in mr-ui.

---

Closes #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Changelog
 
 ## HEAD
 
+- [breaking] Rewrites `src/components/control-switch` as a functional component.and uses [`swtich` from Radix](https://radix-website-git-release.modulz-deploys.com/docs/primitives/components/switch). 🚨 **Drop options**:
+    - `optional`: Additional text like "optional" should be applied by the consumer using the label prop.
+    - `initialValue` and `validator` which were previously provided by the form component.
 - [breaking] Rewrites `src/components/icon` as a functional component and uses [`accessible-icon` from Radix](https://www.radix-ui.com/docs/primitives/utilities/accessible-icon) as an internal primitive.
 - [breaking] Rewrites `src/components/control-wrapper` as a functional component.
 - [feature] Introduces TypeScript to the source code and converts:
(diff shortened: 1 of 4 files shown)


src/docs/tmp-mbx-assembly-candidates.css
```diff
@@ -0,0 +1,152 @@
+/* TODO!! This is a temporary file until https://github.com/mapbox/mbx-assembly/pull/62 is in mbx-assembly and deployed */
+
```
tristen, line 2: I'm adding in this file as a temporary measure until we rule what we need for mbx-assembly and deploy a new version to CDN. This will come in handy for other components we rewrite.

badiuoanaalexandra (approved):
Everything looks good to me and seems to work as expected.

src/docs/tmp-mbx-assembly-candidates.css
tristen: Resolved in chat but I'll circle back on mbx-assembly once things feel resolved in this file.

Issue #156, opened 2022-06-23
v3


tristen:
Stubbing a placeholder epic for tracking work that rewrites mr-ui

tristen:
Modifying the title. It looks like mr-ui never pulled itself out of `2.0.0-beta.2` https://github.com/mapbox/mr-ui/blob/6a9074563941dec6ac7a0c4728f777bd34a8adf9/package.json#L4 so this should likely be a `2.0.0` release.

PR #171, opened 2022-07-06
Drop Underreact for CRA

### LINKED ISSUE #170: Drop Underreact
See https://github.com/mapbox/mr-ui/issues/168#issuecomment-1176489528 for context. I have a working branch that uses create-react-app so let's go with that!

tristen:
Drops a lot of dev depenencies in favor of [create-react-app](https://create-react-app.dev/). Note that as part of this change, I needed to shuffle tests outside of `__tests__` as the Jest test pattern matching in CRA was throwing an error that 0 tests were found on files matching `{component}-test-cases.js`. [CRA apparently allows you to configure this](https://create-react-app.dev/docs/running-tests/#configuration) but it _seems_ as though it only allows you to append existing patterns to the default.

### Reviewer checklist

- [x] Tests run
- [x] Docs generate and run
- [x] npm script commands continue working as expected

---
Closes #170


package.json
```diff
@@ -1,42 +1,19 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "3.0.0-alpha.1",
+  "version": "2.0.0-beta.3",
```
tristen, line 3: See https://github.com/mapbox/mr-ui/issues/156#issuecomment-1175696978. We never got to a major 2.0.0 release so modifying this to the next beta version.

src/components/icon/__tests__/icon.test.tsx
```diff
@@ -66,9 +66,9 @@ describe('inline', () => {
       inline: true
     };
 
-    const { baseElement, getByTestId } = render(<Icon {...props} />)
+    const { baseElement } = render(<Icon {...props} />)
     expect(baseElement).toMatchSnapshot();
-    expect(getByTestId('icon-check')).toHaveClass('inline-block');
+    expect(screen.getByTestId('icon-check')).toHaveClass('inline-block');
```
tristen, line 71: Apparently eslint prefers using `screen` for queries which came up after running tests so I modified this test.

src/components/popover/calculate-position-context.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-unused-expressions */
```
tristen, line 1: This eslint error came up too which I'm just going to ignore outright as we will be replacing this.

src/components/popover/popover-positioner.js
```diff
@@ -135,22 +135,26 @@ class PopoverPositioner extends React.PureComponent {
     const bodyElement = this.bodyElement;
     if (bodyElement) {
       bodyElement.style.setProperty('visibility', 'visible');
-      for (let key in positionData.bodyPositioningStyle) {
-        bodyElement.style.setProperty(
-          key,
-          positionData.bodyPositioningStyle[key]
-        );
+      if (positionData) {
```
tristen, line 138: Tests were failing without this conditional check. Without dumping too much more time into why ... I think this is an OK change as we will be replacing popover-positioner with https://www.radix-ui.com/docs/primitives/components/popover

src/components/copiable/copiable.test.js
```diff
@@ -50,7 +53,7 @@ describe(testCases.basic.description, () => {
     expect(select).toHaveBeenCalledWith(mockSelectableEl);
   });
 
-  test('shows copy hint when focused', () => {
+  test.skip('shows copy hint when focused', () => {
```
tristen, line 56: I'm skipping this snapshot for now but plan on returning to it when this component is re-written. I've dumped too much time into why there's some flakiness fetching the element from `ref` which is preventing the text from rendering in the component.

.gitignore
```diff
@@ -1,9 +1,23 @@
...
+/src/docs/data
+
+# misc
 .DS_Store
-_site*
-src/docs/data
-pkg
-src/test-cases-app/component-index.js
-dist
+.env.local
```
badiuoanaalexandra, line 16: Are these having a very specific config that cannot be committed or what's the reason why these are ignored? In the past I had these as part of the repo if the config was the same for everyone.

badiuoanaalexandra:
Seems that there is a file missing? Ran npm ci and then ran start and build-docs and I get 
image
Also I tried build and I get this 
image

.gitignore
badiuoanaalexandra: or actually do these even exist at all, since this is a library?

tristen: Yeah ... so this is sort of boilerplate after initially running `create-react-app . --template typescript`

badiuoanaalexandra:
Seems to work nicely now, the only one I haven't checked is the deploy-docs. I was afraid it will replace existing docs somewhere live :)

badiuoanaalexandra (approved):
Very nice that start now actually opens the browser and you can see the real generated files!

PR #175, opened 2022-07-11
Rewrite modal

### LINKED ISSUE #172: Rewrite modal component

### Acceptance Criteria

- TypeScript
- Functional component
- Radix inner workings. Here's the API Reference for dialog: https://www.radix-ui.com/docs/primitives/components/dialog
- We might want to additionally support the behavior of Radix's [alert dialog](https://www.radix-ui.com/docs/primitives/components/alert-dialog) in this component.
- Make the alert dialog dismissible or not dismissible through a separate prop

tristen:
This PR rewrites `src/components/modal` as a functional, typescript component.and uses [`DialogPrimitive ` from Radix](https://www.radix-ui.com/docs/primitives/components/dialog#api-reference). 

🚨 **Breaking `button` prop changes**:
- `component` which accepted a component or a string equally a DOM value. This was a little problematic to type in TypeScript and seems a little questionable to have in the first place. I think this is one we can rule on including after audting usage across Mapbox.
- Renamed `discouraging` to `tertiary`

🚨 **Breaking `modal` prop changes**:
- `alert` this wasn't being used anywhere 
- `focusTrapPaused` this wasn't used anywhere and isn't exposed as an option from Radix

---

Closes #172


src/components/modal/modal.tsx
```diff
@@ -1,13 +1,36 @@
-import React from 'react';
+import React, { ReactElement, ReactNode, forwardRef, CSSProperties } from 'react';
```
badiuoanaalexandra, line 1: forwardRef is not in use, so can be removed

src/components/modal/modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      };
+      const { rerender } = render(<Modal {...props} />);
+      expect(screen.getByTestId('event-trap')).toBeInTheDocument();
+      rerender(<Modal {...props } allowEventBubbling={true} />);
+      expect(screen.queryByTestId('event-trap')).not.toBeInTheDocument();
+    });
+  });
+
+  describe('basic, large', () => {
+    const mockedOnExit = jest.fn();
```
badiuoanaalexandra, line 43: mockedOnExit not used, so can be removed. I propose if possible to have our pipeline also fail on eslint warnings, so that we can fixed these by default? What do you think?

src/components/modal/modal.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    test('fires onExit', () => {
+      render(<Modal {...props} />);
+      fireEvent.click(screen.getByTestId('modal-close'));
+      expect(mockedOnExit).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('all options', () => {
+    const mockedOnExit = jest.fn();
```
badiuoanaalexandra, line 97: mockedOnExit not used, so can be removed. I propose if possible to have our pipeline also fail on eslint warnings, so that we can fixed these by default? What do you think?

tristen:
> Luckily in this case if you scroll up and down you are able to see the full content, so I don't think it's anything urgent, but should be considered.

Yeah, this maintains the current behavior of the modal today so we might want to keep it at least for now. Setting `padding: "none"` and customizing the modal content outer div with max height + overflow can also achieve that desired result today.

src/components/modal/modal.tsx
tristen: Confirmed it's used in this file.

tristen: Oh whoops it's not! and the extended eslinit rule caught it 😅. I was looking at a rebase commit which still had it.

PR #177, opened 2022-07-15
Rewrite accordion

### LINKED ISSUE #174: Rewrite controlled-accordion component

- TypeScript
- Functional component
- Radix inner workings. Here's the API Reference for accordion which can be controlled or uncontrolled: https://www.radix-ui.com/docs/primitives/components/accordion#api-reference

tristen:
This PR creates `src/components/accordion` as a functional, typescript component.and uses [`AccordionPrimitive ` from Radix](https://www.radix-ui.com/docs/primitives/components/accordion#api-reference). 

🚨 **Breaking changes**:
- `accordion-item`, `accordion-controlled`, `accordion-uncontrolled` is replaced by a single `accordion` component that requires control.
- All props for controlled-accordion remain the same!

---

Closes #174


src/components/accordion/accordion.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <AccordionPrimitive.Root type="single" value={activeItem} onValueChange={onToggle} collapsible>
+      {items.map(renderItems)}
+    </AccordionPrimitive.Root>
+  );
+}
+
+Accordion.propTypes = {
+  items: PropTypes.arrayOf(
```
badiuoanaalexandra, line 65: The description for items is missing

badiuoanaalexandra (approved):
Approving with suggestion to add description for items prop. This didn't exist before either, but it would nice to have a description. Otherwise the accordion works as expected.

Issue #116, opened 2021-02-23
`themItemContainer ` prop is not used in the tablist


shabnomnom:
Per documentation https://mapbox.github.io/mr-ui/#tablist `themItemContainer ` is a valid prop for the tablist, but I was not able to utilize this prop. Looking closer looks like `themItemContainer ` is never rendered beyond the initial definition. `themItemContainer ` https://github.com/mapbox/mr-ui/blob/d2e0a19caa9790325953ba3487b94cfae2153b19/src/components/tab-list/tab-list.js#L40

tristen:
Closing as we'll be removing this as part of #180

PR #181, opened 2022-07-22
Issue-180 Create Tabs component using radix tabs

badiuoanaalexandra:
This PR deletes tab-list component for the following reasons:
- not possible to take advantage of the radix tabs out of the box functionality (such as key bindings) if we need to keep the functionality of rendering popover for mobile tabs
- from the audit we discovered that tab-list was not used anywhere

This PR replaces underline-tabs component with tabs. Almost all the functionality underline-tabs was migrated to radix, with a few exceptions:
- tabs don't support links - unfortunately I discovered that the key bindings do not work if we don't allow radix to create it's own buttons (radix will create buttons for the tabs if asChild is not present in the trigger). I also think that if links are involved then those should be something like "navigation" component and not tabs. I don't really see a place where navigation and section tabs would be used in the same component
- some prop names were changed just to avoid using confusing names
- possibility to add tab content that is automatically rendered with the active tab has been added to the component
- due to the way radix works I had to add a prop called **themeTabsContainer** that renders special css classes around the tab triggers and doesn't affect the content

Radix tabs supports out of the box functionality for switching tabs, unfortunately due to the way mr-ui is implemented at the moment I couldn't add this. mr-ui doesn't include any css of its own atm, the css comes from assembly package. 
In order to support more radix functionality I think mr-ui should have some css of it's own, however this seems like a bigger change to the way we work and so would need to discuss with the whole team. 
Other props that radix supports for tabs weren't added because our current css wasn't supporting it.


src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,141 @@
...
+import * as RadixTabs from '@radix-ui/react-tabs';
+import { TabItemProps, SizeType } from './types';
+/**
+ * For navigation between sections. Tabs will be always buttons
+ *
+ */
+
+interface Props {
+  tabs: Array<TabItemProps>;
+  activeTab?: string;
```
miafan23, line 15: Regarding the original implementation where `active` is a required prop, I think `activeTab` here should be a required prop too. Considering our new `tabs` component is including tab content now, the height of this component changes suddenly when the activeTab is set to a value from null.

miafan23:
When you hovering over the space between two tabs, there is no tab is highlighted. However, when you click the empty space, the later tab is activated. 
For the original underline-tabs component, the empty space is not clickable. For the example on the Radix website, when you hovering on the empty space, the according tab would be highlighted. Both of above experience are better than current implementation from my point of view.

https://user-images.githubusercontent.com/9078950/180781948-c553a046-797e-4b9b-afbd-d3c77f191e83.mp4

miafan23 (commented):
image
Maybe add a padding to the content in the examples to make them looks better and more professional.

tristen:
Leaving a note to also address https://github.com/mapbox/mr-ui/issues/116 as part of this PR.

miafan23 (approved):
Great changes! LGTM!

Issue #179, opened 2022-07-19
Remove components that are not needed


badiuoanaalexandra:
According to the mr-ui audit document and how we want our library to be used, there are a few components that can be removed:

- **badge** - same behavior can be achieved with tooltip and perhaps a cssClass for the trigger if we want them all to look the same
- **control-card** - not used anywhere
- **control-legend** - not used anywhere, doesn’t seem to do anything special. I think having control-label is enough
- **go-link** - I don’t think a link should be a component, but just have default css for links 
- **heading** - same as link I think we should have default styles for the headings and not a component for them
- **new-tab-link** - normal link should be used for this, with default styling that just opens a new tab. No need for a component

The list should be discussed with @tristen and @VLuisa

tristen:
Going to quickly take care of this one and send over to @VLuisa for review.

tristen:
Closed per https://github.com/mapbox/mr-ui/pull/183

PR #183, opened 2022-07-28
Remove components

tristen:
Drop components as deemed underutilized or no longer needed per https://github.com/mapbox/mr-ui/issues/179


src/components/control-radio-set/control-radio-set.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import omit from '../utils/omit';
-import ControlLegend from '../control-legend';
```
tristen, line 4: Some of our control group elements optionally used `ControlLegend` so they have been dropped from these maintained components along with the supported props related to it.

VLuisa (commented):
Overall this looks good!

@tristen I don't feel super strongly about this but just wanted to say I’m not entirely convinced we should get rid of ‘Badge’ - we do use the [badge pattern](https://github.com/mapbox/studio/blob/9ac1fa9e2b8f998f6dca86a7795c2ad0c2efb460/src/shared_components/announcement_item_tag.js) in Studio, we just haven’t made use of mr-ui for it and I do think it could be much cleaner to have the component than having to manually recreate all the css classes for every tag color combination (correct me if that’s not what we’d have to do). 

image

tristen:
oh yeah @VLuisa that's a good point to raise. I could be convinced to keep it. I was thinking the CSS classes used to create each badge could be added to https://github.com/mapbox/studio/blob/main/src/util/ui_util/get_style.ts which is a more shareable place within Studio to maintain consistency.

VLuisa:
@tristen hmm i think i actually like that more. And if it's in Studio then we coulddd be more opinionated than in mr-ui and use a semantic naming structure (ex: `badgeNew`, `badgeBeta`, etc), kind of like they're set up in this tag component... 
```
'bg-orange-faint color-orange': type === 'deprecated',
'bg-green-faint color-green': type === 'new',
'bg-blue-faint color-blue': type === 'beta',
'bg-pink-faint color-pink': type === 'ui'
```

PR #184, opened 2022-08-03
Rewrite Popover, drop popover trigger

### LINKED ISSUE #173: Rewrite popover component

- TypeScript
- Functional component
- Radix inner workings. Here's the API Reference for dialog: https://www.radix-ui.com/docs/primitives/components/popover#api-reference

tristen:
This PR rewrites `src/components/popover` as a functional, typescript component and uses [`PopoverPrimitive` from Radix](https://www.radix-ui.com/docs/primitives/components/popover#custom-apis) under the hood. 

🚨 **Breaking changes**:
- Dropped the separate uncontrolled PopoverTrigger component in favor of composition and promoting controlled implementations.
- Dropped`popover` prop changes:
  - `getAnchorElement` in favour of an exported PopoverTrigger component instead which Radix encourages
  - `receiveFocus` which was a boolean (`true` by default) telling the popover whether or not to be in focus. This can now be achived by e.preventDefault(); in `getInitialFocus`.
  - `trapFocus` unused property
  - `ignoreClickWithinElement` unused property
  - `getContainingElement` unused property
  - `onElement` unused property
  - `zindex` unused property
  - `observeSize` not exposed as an option outside of Radix and an unused prop after evaluating implementation
  - `containWithinViewport` unused property and inclined to follow Radix default behaviour

---

Closes #173


tristen:
I'm running into a similar issue with tests as [reported here](https://github.com/radix-ui/primitives/issues/1608) so I'm going to pause on this work for a bit.

andy-hook:
> I'm running into a similar issue with tests as [reported here](https://github.com/radix-ui/primitives/issues/1608) so I'm going to pause on this work for a bit.

Hey @tristen I posted a follow up [here](https://github.com/radix-ui/primitives/issues/1608#issuecomment-1210381690) if it's helpful.

src/components/copiable/copiable.test.js
```diff
@@ -54,16 +54,18 @@ describe(testCases.basic.description, () => {
...
       mockEvent
     );
     wrapper.update();
     expect(wrapper.find('Popover').length).toBe(1);
+    /*
     return delay(FEEDBACK_TIME + 100).then(() => {
       wrapper.update();
       expect(wrapper.find('Popover').length).toBe(0);
     });
+    */
```
miafan23, line 68: Maybe add some comments to explain why those test code is commented out? Then when we are rewriting `copiable` component, we'll be more confident to uncomment them or remove them.

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,180 @@
...
+      </PopoverPrimitive.Anchor>
+    );
+  });
+
+  const getContent = () => {
+    if (typeof content === 'function') {
+      return content();
+    } else {
+      return content;
+    }
```
miafan23, line 75:
```suggestion
    if (typeof content === 'function') {
      return content();
    }
    return content;
```
The `else` seems unnecessary here

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,180 @@
...
+        {hasPointer && <PopoverPrimitive.Arrow width={6} height={6} fill={fill} />}
+      </PopoverPrimitive.Content>
+      <Anchor />
+    </PopoverPrimitive.Root>
+  );
+}
+
+Popover.propTypes = {
+  /**
+   * The trigger eleemnt.
```
miafan23, line 103: A typo :stuck_out_tongue_closed_eyes:

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,180 @@
...
+   * to remove the Popover from the rendered page.
+   */
+  onExit: PropTypes.func,
+  /**
+   * A function that returns a DOM node. Use to specify which element should
+   * receive focus when popover is first rendered.
+   */
+  getInitialFocus: PropTypes.func,
+  /**
+  /**
```
miafan23, line 172: Extra comment starter

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+Popover.propTypes = {
+  /**
+   * The trigger eleemnt.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * The popover content. This can either be a string, valid JSX, or a function
+   * returning either.
+   */
```
miafan23, line 109: Missing document for `active` prop

miafan23 (changes requested):
Looks great!
Some changes to the document are required and I'll paste a few videos about some affected UI that I am not able to get the reasons from the code.

miafan23:
The button flashes when you click it. And the content of popover flashes in the last example. I checked the original mr-ui and radix UI official examples, seems both of them don't have this issue.



https://user-images.githubusercontent.com/9078950/184605886-f069cece-cf45-4f91-8574-0c6aa067c30c.mp4

miafan23:
Since you touched `copyButton` component, so I also took a look for its example page and found the height of the wrapper changes when you click the copyButton.

https://user-images.githubusercontent.com/9078950/184606429-dd6d1e4c-8762-4903-85b6-773eb8b41e23.mp4

src/components/copiable/copiable.test.js
tristen: 👍 These were re-written in https://github.com/mapbox/mr-ui/pull/186

src/components/popover/popover.tsx
tristen: Ah! good catch

tristen:
@miafan23 this is ready for another review!

miafan23 (approved):
Excellent!

PR #192, opened 2022-08-16
Rewrite control-label

### LINKED ISSUE #189: Rewrite control-label component
### Criteria:

- TypeScript
- Functional component
- Radix inner workings. Here's the API Reference for label: https://www.radix-ui.com/docs/primitives/components/label#api-reference

tristen:
Simple changes to the control-label component to use TypeScript, React testing library, and accessibility enhancements provided by [`@radix-ui/react-label`](https://www.radix-ui.com/docs/primitives/components/label#api-reference).

---

Closes #189


src/components/tabs/index.ts
```diff
@@ -1,3 +0,0 @@
-import main from './tabs';
-
-export default main;
```
tristen, line 3: Just noticed this unnecessary file as `src/components/tabs/index.js` exists

src/components/icon/icon.test.tsx
```diff
@@ -8,8 +8,8 @@ jest.mock('../utils/get-window');
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const getWindowMock = getWindow as any;
 
-describe('renders', () => {
-  test('basic', () => {
+describe('basic', () => {
+  test('renders', () => {
```
tristen, line 12: Small naming change to an unrelated component. I noticed the two were in the wrong order.

src/components/control-label/examples/control-label-example-basic.tsx
```diff
@@ -0,0 +1,14 @@
+/*
+Basic.
+*/
+import React from 'react';
+import ControlLabel from '../control-label';
+
+export default function Example() {
+  return (
+    <>
+      <ControlLabel id="name" text="Enter your name" />
```
miafan23, line 10: Maybe replace the value of `id` with another one? It is the same as the id of `controlFile`, so when I click this label on the example page, the dialog for selecting files popped up.

miafan23 (approved):
LGTM! Only one minor fixing might be required.

src/components/control-label/examples/control-label-example-basic.tsx
tristen: ohhh nice catch! I might change this in https://github.com/mapbox/mr-ui/pull/186 then as it seems strange for id to be `name` there.

PR #193, opened 2022-08-16
Rewrite icon-text

### LINKED ISSUE #190: Rewrite icon-text component
### Criteria:

- TypeScript
- Functional component
- _possible_ radix inner workings or adoption of components re-written with radix like icon and control-label

tristen:
Simple changes to the icon-text component to use TypeScript, and React testing library. **Note** that I removed the `gap` which was not actually being used but documented as an option.

---

Closes #190


miafan23 (approved):
LGTM!

PR #195, opened 2022-08-18
Rewrite control-select

### LINKED ISSUE #178: Rewrite control-select using radix select
- Rewrite control-select component using [radix react-select](https://www.radix-ui.com/docs/primitives/components/select)
- Typescript should be used
- Keep as much as possible of the current functionality

Below you can find information about current usage of control-select. 

**control-select**
**Used in**: mapbox/install-atlas, mapbox/admin, mapbox/studio, mapbox/help, mapbox/account, mapbox/search-lab, mapbox/www.mapbox.com, mapbox/dr-ui

**Props used**: id, value, label, themeLabel, onChange, options, themeControlSelect, disabled, themeControlWrapper, themeControlSelectContainer, data-test, getControlProps(‘country’), aside, validationError

There are repos that might be used on mobile phone, so we should also check how the select works on mobile. 

Taken from mr-ui audit document https://docs.google.com/document/d/1x-fCsbmvcsYV2Y49cC2eFiV_XonECpOnkYAkGBtcNIM/

tristen:
Simple changes to the control-select component to use TypeScript, and React testing library.

---

Closes #178


miafan23 (commented):
For these examples, would it be better to set the selected option to the value? 
The rewritten component is good!

miafan23 (approved):

PR #199, opened 2022-10-04
Support disabled state in modal actions

tristen:
To address downstream usage in Studio


AlexanderBelokon (approved):
LGTM!

BTW, is it possible to add comments to individual `expect`s like we can in `tape` and not bail out of the test case after the first `expect` failure? I think that makes testing multiple different characteristics of a result after a single setup much more useful.

PR #202, opened 2022-10-07
Remove children prop from toast component

### LINKED ISSUE #201: Allow the `children` prop to be optional
I have a use case where I want to broadcast from the toast component after the onExit callback of another component and it would be nice if I didn't have to pass children like so:

```jsx
import React, { useState } from 'react';
import Toast from '@mapbox/mr-ui/toast';
import CustomComponent from './custom-component';

function myProp() {
  const [done, setDone] = useState(true);
  return (
    <>
      <CustomComponent onExit={() => setDone(true)}
      <Toast active={done} />
    </>
  );
}
```

Considering this is a controlled component, I wonder if we should just omit a children prop altogether. Thoughts @VLuisa?

VLuisa:
Closes #201 

@tristen for review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## HEAD
 
+- [fix] Remove `children` prop from **Toast** component.
+
 ## 2.1.0
 
 - [feature] Add **Toast** component.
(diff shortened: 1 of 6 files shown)


tristen (approved):
LGTM!

PR #204, opened 2023-01-27
Support a JSX for a control file select label

tristen:
- Small markup simplification to the icon-text component
- Support `ReactNode` as a `displayLabel` prop in file-select to properly handle scenarios where you might want to truncate a long label name.


miafan23 (approved):
LGTM!

PR #205, opened 2023-02-06
Add `children` and `themeTooltip` props to icon-button

tristen:
Adding additional props to the copy-button to unblock react-colorpickr using it for a customized copiable interface


AlexanderBelokon (approved):
LGTM!

PR #206, opened 2023-03-15
Add context menu component

tristen:
This PR:

- Updates Radix packages to their latest.
- Adds a context menu component powered by the [API provided in Radix](https://www.radix-ui.com/docs/primitives/components/context-menu#api-reference).


tristen:
Just flagging that the diff of changes looks big on account of snapshots changing with radix package updates. The implementation of the context menu + tests is like half of it.

src/components/context-menu/context-menu.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+type Options = {
+  title?: string;
+  options: Array<Option>;
+}
+
+type Option = {
+  label: ReactNode;
+  value: string;
+  keyCode?: string;
```
VLuisa, line 15: Would we expect someone to use the `<KeybindingTooltipWrapper />` with this in Studio? Or is that redundant?

src/components/context-menu/context-menu.tsx
```diff
@@ -0,0 +1,151 @@
...
+  themeItem?: string;
+  themeTitle?: string;
+  themeSeparator?: string;
+}
+
+export default function ContextMenu({
+  children,
+  onChange,
+  options,
+  themeControlWrapper = 'bg-white round shadow-darken10 px6 py6 txt-s wmin240',
```
VLuisa, line 34: Nitpick: For the default values, the `shadow-darken10` feels odd on the disabled line item b/c it hints at interactivity that's not available

src/components/context-menu/context-menu.tsx
```diff
@@ -0,0 +1,151 @@
...
+  themeItem = 'px6 py6 round color-blue-on-hover bg-gray-lighter-on-hover',
+  themeTitle = 'px6 py6 txt-bold',
+  themeSeparator = 'border-t border--gray-light my6'
```
VLuisa, line 37: having this in the api is nice 👍

src/components/context-menu/context-menu.tsx
VLuisa: `txt-s` felt a bit small for the list item labels, what do you think about `txt-ms` instead?

src/docs/index.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './app';
 import './prism-theme.css';
+import './miscellaneous.css';
```
VLuisa, line 5: What's this for?

VLuisa: Oh duh, scrolling down to the actual file 😅

src/components/context-menu/context-menu.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+ContextMenu.propTypes = {
+  /**
+   * The trigger element.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * An array of objects that represent each option group.
+   * Each object contain two fields:
```
VLuisa, line 113:
```suggestion
   * Each object contains two fields:
```
typo

VLuisa (changes requested):
This is really exciting! I think the overall style looks great for use in Studio and I like the flexibility for styling you're providing in the api. Left a few small comments for your thoughts.

src/components/context-menu/context-menu.tsx
tristen: ohhh that's a great call! so instead os a string this should accept a `ReactNode`?

src/docs/index.js
tristen: It is ... It _might_ actually be worth adding these specific rules to mbx-assembly instead 🤔

src/components/context-menu/context-menu.tsx
tristen: > Nitpick: For the default values, the `shadow-darken10` feels odd on the disabled line item b/c it hints at interactivity that's not available

I'm not sure I understand this one. Could you share a screenshot?

VLuisa: @tristen yep! I don't think the disabled item should have a background color on hover
image

src/components/context-menu/context-menu.tsx
VLuisa: probably? may be something we have to play with as we implement it in Studio.

src/docs/index.js
VLuisa: could be! are they reusable in other components for focus states?

samanpwbb:
Wow

VLuisa (approved):
LGTM!

PR #208, opened 2023-03-31
Change shadow color on dark tooltip

VLuisa:
Added a light shadow to the dark tooltip to improve visibility against dark backgrounds:
image

Also updated all tooltip shadows to use `drop-shadow` for more accurate shading.

@tristen for approval

### DIFF
--- a/src/components/control-text/__snapshots__/control-text.test.tsx.snap
+++ b/src/components/control-text/__snapshots__/control-text.test.tsx.snap
@@ -99,7 +99,7 @@ exports[`ControlText all options renders as expected 1`] = `
     style="position: fixed; left: 0px; top: 0px; transform: translate3d(0, -200%, 0); min-width: max-content;"
   >
     <div
-      class="bg-gray-dark border--gray-dark color-white border shadow-darken25 round px12 py6 border border-1 border--border--gray-dark"
+      class="bg-gray-dark border--gray-deep color-white border shadow-darken25 round px12 py6 border border-1 border--border--gray-deep"
       data-align="center"
       data-side="top"
       data-state="open"
@@ -113,7 +113,7 @@ exports[`ControlText all options renders as expected 1`] = `
         style="position: absolute; bottom: 0px; transform: translateY(100%); visibility: hidden;"
       >
         <svg
-          fill="#0E2127"
+          fill="#2E4E5D"
           height="6"
           offset="6"
           preserveAspectRatio="none"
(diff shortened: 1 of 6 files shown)


tristen:
Ah! @VLuisa one thought: is it possible to use a `stroke` attribute to the triangle element? https://github.com/mapbox/mr-ui/blob/66a47401b684a582f6800bbcb8176052d4cebf87/src/components/tooltip/tooltip.tsx#L91

I wonder if that might achieve the clean outline around the whole tooltip as you had proposed.

VLuisa:
@tristen Good point! If I'm understanding correctly, I think that only gets it partway there? Because then each item has it's own stroke (illustrated with red here) as opposed to having one trace all the way around (which can be done but kind of in a hacky way which I'm not sure is worth it?)
image

tristen:
Ah! @VLuisa thanks for trying that out! I wish `stroke` accepted positional color values 😞 I can _almost_ achieve it using the [`drop-shadow` filter property](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow) but it looks a little funky close-up.

Screenshot 2023-04-01 at 11 00 14 AM


Here's the CSS for that 

```css
{ 
  filter: drop-shadow(0 1px 0 red) drop-shadow(0 -1px 0 red) drop-shadow(1px 0 0 red) drop-shadow(-1px 0 0 red)
}
```

VLuisa:
@tristen Ooh nice! This feels like it may work because it'd never be observed as closely as in a large screenshot? 
Just to confirm the workflow, would this require adding a new CSS class in assembly?

tristen:
> Just to confirm the workflow, would this require adding a new CSS class in assembly?

I think this can be added as inline CSS because it's pretty specific to this option and doesn't require pseudo-selectors. Something like this? 

```js
import variables from '@mapbox/mbx-assembly/dist/variables.json';

return (
  <div style={{
    filter:  `
      drop-shadow(0 1px 0 ${variables['--gray-deep']})
      drop-shadow(0 -1px 0 ${variables['--gray-deep']})
      drop-shadow(1px 0 0 ${variables['--gray-deep']})
      drop-shadow(-1px 0 0 ${variables['--gray-deep']})
    `
  />
);

```

VLuisa:
@tristen This (mostly) worked for the dark tooltip - as you said, if you don't look too closely. 
image

So naturally, I tried applying it to all the tooltips and... it doesn't look great 😅 
image
image


I'm leaning towards keeping the dark tooltip as it is in this PR so that they're all consistent. But we could also use this outline styling for _just_ the dark tooltip - curious what you think!

VLuisa:
@tristen can't re-request review through the github UI, but I made our proposed changes using `drop-shadow`!

tristen (approved):
I think this looks great!

PR #214, opened 2023-08-30
add margin prop to Modal

chriswhong:
After recent docs redesign work, a new fixed header is in place throughout all docs sites.  The `Modal` component's vertical margin is not large enough to compensate for the height of the header and the top part of the modal is obscured.

This PR adds a `margin` prop to the `Modal` component.  When set to `large`, it will add the `my120` assembly class, giving enough clearance to the modal.  

Before
image

After
image

### DIFF
--- a/src/components/modal/modal.tsx
+++ b/src/components/modal/modal.tsx
@@ -13,6 +13,7 @@ interface Props {
   accessibleTitle: string;
   size?: 'small' | 'large' | 'auto';
   padding?: 'large' | 'none';
+  margin?: 'large' | 'default';
   onExit?: () => void;
   allowEventBubbling?: boolean;
   initialFocus?: string;
@@ -50,6 +51,7 @@ export default function Modal({
   accessibleTitle,
   size = 'large',
   padding = 'large',
+  margin = 'default',
   allowEventBubbling = false,
   initialFocus,
   primaryAction,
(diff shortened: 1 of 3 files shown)


src/components/modal/modal.tsx
```diff
@@ -109,7 +116,7 @@ export default function Modal({
     onOpenAutoFocus?: (e) => void
   } = {
     className: classnames(
-      `relative my12 my60-mm ${widthClass} bg-white round`,
+      `relative ${marginClass} ${widthClass} bg-white round`,
```
tristen, line 119:
```suggestion
      `fixed hmax-viewport-11/12 ${widthClass} bg-white round`,
```

@chriswhong what do you think about using `fixed` instead and also adding the following CSS to the element itself.

```js
style: {
  top: '50%',
  left: '50%',
  transform: 'translate(-50%, -50%)'
}
```

That should center the modal and let us avoid adding fixed margins to begin with.

chriswhong: I lean towards modals opening near the top of the screen as a sensible default.  This component could be used for content that is taller than the viewport, and the proposed change would cut off the top and bottom in that situation.  (maybe for that much content we should not use a modal, but it's a footgun we should consider.

tristen: > This component could be used for content that is taller than the viewport

Oh! That was the intention of adding `hmax-viewport-11/12` so the modal is always shown but we would favor using an overflow scroll for content that extends beyond the viewport.

tristen (approved):
Approving to unblock as the implementation looks 👌

PR #216, opened 2023-09-18
replace calls to getWindow with a simpler check

### LINKED ISSUE #210: Calls to getWindow() prevent successful use in SSR
I recently attempted to update to the latest `mr-ui` in one of the Mapbox docs sites, and kept getting a `window not available` error during build.

After some digging, I traced it to the `Icon` component, which calls the util `getWindow()`.  `getWindow()` throws an error if `window` does not exist, making it impossible to render `Icon` in an SSR build.

https://github.com/mapbox/mr-ui/blob/main/src/components/icon/icon.tsx#L32

cc/ @tristen, I am curious what you would recommend for using mr-ui in an SSR environment.  `Icon` is used heavily across many of our docs sites.

chriswhong:
Removes the util `getWindow()` which was causing issues with SSR in docs sites (closes #210)


tristen (approved):
LGTM!

PR #218, opened 2024-01-19
Custom select UI

tristen:
- Dependent on https://github.com/mapbox/mbx-assembly/pull/68
- Closes https://mapbox.atlassian.net/browse/STUDIO-2018


AlexanderBelokon (changes requested):
This is a sweet component! 🔥

I only found one issue with it on the examples page: when filtering, sometimes the input loses the focus. It seems to happen every time the currently selected item is added or removed from the current list, but does not happen otherwise.

Curiously enough, it only happens in Chrome, not in Firefox 🤔

Is this a problem of the example or of the component?

https://github.com/mapbox/mr-ui/assets/2198527/c2c2eda9-f00e-4185-ad28-40fb4b02ea49

tristen:
@AlexanderBelokon thanks for flagging 👍 Looks like a [related/possible upstream issue](https://github.com/radix-ui/primitives/issues/2193) but I hacked a little fix for it on the demo page.

AlexanderBelokon (approved):
Works like a charm! 🧙

PR #224, opened 2024-07-17
Fix vulnerabilities, remove cpy lib and upgrade to node 21

badiuoanaalexandra:
- Fixed vulnerability related to braces 
- removed cpy lib - cpy had also vulnerability related to braces lib. cpy has many issues and shouldn't be used anymore, so removed it and replaced the functionality with node native libs (fs and path).
- updated to node 21
- tested all scripts and everything works

Since this is just a dev change I didn't update any packages.


tristen (approved):
LGTM

.nvmrc
```diff
@@ -1 +1 @@
-16
\ No newline at end of file
+21
```
AlexanderBelokon, line 1: I don't think we should ever use non-LTS Node releases, and all odd ones don't get LTS

badiuoanaalexandra: Aaah.. that's annoying.. I didn't know that odd ones don't get it. I'll change to 20

badiuoanaalexandra: https://github.com/mapbox/mr-ui/pull/225

PR #227, opened 2024-11-21
Add onCopy property to Copiable component

badiuoanaalexandra:
Adds possibility to pass onCopy to Copiable component


miafan23 (approved):
LGTM

PR #228, opened 2024-12-02
Support configuring the click behavior on underlay of Modal

miafan23:
Added a new prop `exitOnUnderlayClicked` for Modal. When it is `false`, a click on the underlay will not close the modal


badiuoanaalexandra (approved):
Works as expected! 🚀

PR #230, opened 2024-12-04
Support configuring z-index for modal

miafan23:
Add a `zIndex` prop for modal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "2.13.0",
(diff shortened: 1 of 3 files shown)


tristen (approved):

tristen:
@miafan23 I'm curious, what changed to require this? I thought radix dialogs had hardcoded z-index at the highest value

miafan23:
> I'm curious, what changed to require this? I thought radix dialogs had hardcoded z-index at the highest value

Seems not. It has no z-index value. 
image

PR #231, opened 2024-12-05
Support configuring z-index for popover

miafan23:
When popover is in a container with a z-index, it has to have a z-index to be visible


badiuoanaalexandra (approved):
LGTM

PR #234, opened 2025-02-18
A new component Drawer

miafan23:
Created a new component Drawer based on radix/dialog component

image


src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  themeContent = '',
+  accessibleTitle = 'drawer',
+  onExit
+}: Props): ReactElement {
+  const overlayProps: {
+    className: string;
+    style: CSSProperties;
+    'data-testid': string;
+  } = {
+    className: `fixed top bottom left right bg-darken50 ${themeOverlay}`,
```
tristen, line 42: Should stylistic attributes like background darkening be the default value of `themeOverlay`?

src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,129 @@
...
+            {children}
+          </DialogPrimitive.Content>
+        </DialogPrimitive.Overlay>
+      </DialogPrimitive.Portal>
+    </DialogPrimitive.Root>
+  );
+
+  if (!allowEventBubbling) {
+    return <EventTrap>{drawer}</EventTrap>;
+  }
```
tristen, line 91: Nice!

src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  const contentProps: {
+    className: string;
+  } = {
+    className: classnames(
+      `overflow-auto scroll-styled fixed top bottom ${themeContent}`,
+      {
+        'left round-tr round-br': side === 'left',
+        'right round-tl round-bl': side === 'right',
```
tristen, line 63: I'm not sure that rounding makes sense here when the content stretches to the edges of the viewport.

tristen (changes requested):
Tiny pieces of feedback. Otherwise, tests and functionality looks great!

src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const overlayProps: {
+    className: string;
+    style: CSSProperties;
+    'data-testid': string;
+  } = {
+    className: `fixed top bottom left right bg-darken50 ${themeOverlay}`,
+    'data-testid': 'drawer-overlay',
+    style: {
+      zIndex
+    }
```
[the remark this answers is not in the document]
miafan23: Oh, no! I'll remove it. `themeOverlay` is enough

src/components/drawer/drawer.tsx
miafan23: This is from the design. I'll remove it, I think this should be decided by the application

tristen (approved):

PR #235, opened 2025-03-19
Form to allow changes to control values

Ripley48:
### Description
This PR makes changes to the form component so that it allows `onChange` and `handleFormData` functions to make changes to the control values of the form.

With this change the parent component can have custom logic in the `onChange` function to change the values of the form if needed. Also the `handleFormData` can change the forms values for example based on the server response.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "2.17.0",
+  "version": "3.0.0",
   "description": "UI components for Mapbox projects",
   "main": "index.js",
   "homepage": "./",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "3.0.0",
+  "version": "2.18.0",
   "description": "UI components for Mapbox projects",
   "main": "index.js",
   "homepage": "./",
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -152,13 +152,13 @@ export default function Form({
     nextControlValues[controlName] = controlValue;
     if (formState.current !== formStates.preSubmission) {
       checkValidation(formStates.preSubmission).then(() => {
+        if (onChange) onChange(nextControlValues);
         setControlValues(nextControlValues);
       });
     } else {
+      if (onChange) onChange(nextControlValues);
       setControlValues(nextControlValues);
     }
-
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "2.17.0",
+  "version": "3.0.0",
```
miafan23, line 3: This does not look like a major version change

src/components/form/form.tsx
```diff
@@ -152,13 +152,13 @@ export default function Form({
     nextControlValues[controlName] = controlValue;
     if (formState.current !== formStates.preSubmission) {
       checkValidation(formStates.preSubmission).then(() => {
+        if (onChange) onChange(nextControlValues);
         setControlValues(nextControlValues);
       });
     } else {
+      if (onChange) onChange(nextControlValues);
       setControlValues(nextControlValues);
     }
```
miafan23, line 161: The readability of this part code is not very good. Did you try `await checkValidation(formStates.preSubmission)`? If it works, the code could just be simplified as
```
   if (formState.current !== formStates.preSubmission) {
      await checkValidation(formStates.preSubmission);
    } 
    onChange && onChange(nextControlValues);
    setControlValues(nextControlValues);
```

src/components/form/form.tsx
```diff
@@ -183,12 +183,15 @@ export default function Form({
   const submitForm = () => {
     // Promise.resolve because if handleFormData is a promise,
     // we want to handle loading and error states.
-    Promise.resolve(handleFormData(controlValues))
+    const nextControlValues = { ...controlValues };
+    Promise.resolve(handleFormData(nextControlValues))
```
miafan23, line 187: What's the intention of this change?

Ripley48: This change allows the `handleFormData` to change the control values for example based on server response.

src/components/form/form.tsx
Ripley48: Tested and agree the suggestion is cleaner. Now fixed accordingly.

src/components/form/form.tsx
miafan23: I am bit concerned about this change. My experience is it will take in many unknown issues if the parent component is able to change the children's state value.
This is not a controlled component, if you want to the full control on the values showing in the form, I think you should use controlText component.
Or you should use a different component after the server has responded

PR #236, opened 2025-04-17
Studio-3468 Fix clickOutsideCloses property in Popover component

badiuoanaalexandra:
When passing clickOutsideCloses={false} to popover an error was triggered when trying to close the popup or when clicking somewhere outside. This PR fixes this error. 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "2.18.0",
+  "version": "2.17.1",
   "description": "UI components for Mapbox projects",
   "main": "index.js",
   "homepage": "./",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mr-ui",
-  "version": "2.17.0",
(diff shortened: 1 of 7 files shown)


package.json
tristen, line 1: I think this is just a patch? We aren’t introducing anything new, just fixing a bug!

src/components/control-text/__snapshots__/control-text.test.tsx.snap
tristen, line 1: I don’t understand the inclusion of this `z-index: auto` where is it coming from?

tristen (changes requested):
Just small stuff. Otherwise, this looks good2go

package.json
badiuoanaalexandra: Changed to patch!

src/components/control-text/__snapshots__/control-text.test.tsx.snap
badiuoanaalexandra: I have not changed anything regarding to this one.. so I think something to do with radix

tristen (approved):

PR #238, opened 2025-05-08
Restyle Tab

miafan23:
- Size only affects height and margin now. The font-size should be set by the app itself
- New prop tabBorder 
- Style of active, inactive, hovering components is controlled by *-on-acitve and *-on-hover classes

### DIFF
--- a/src/components/tabs/tab-item.tsx
+++ b/src/components/tabs/tab-item.tsx
@@ -1,86 +1,72 @@
 import React from 'react';
 import classnames from 'classnames';
-import {
-  SIZE_SMALL,
-  SIZE_MEDIUM,
-  SIZE_LARGE
-} from './tabs-constants';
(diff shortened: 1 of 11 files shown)


src/components/tabs/tab-item.tsx
```diff
@@ -42,17 +40,17 @@ export default function TabItem({
...
-  }
+  };
 
   const renderBorder = () => {
     if (!active) {
       return null;
     }
     const borderClasses = classnames('absolute bottom left right border-b', {
-      'border-b--2': size !== SIZE_SMALL
+      'border-b--2': size !== SIZE_SMALL && !overlapBorder
```
miafan23, line 50: Keep the border as 1 px if overlapBorder is true

tristen (approved):
Big fan! and love that prop options have been reduced to 8 as part of this change 👏

PR #251, opened 2026-05-28
Support a custom label display for ControlSelect

tristen:
Moving a version of this from Studio because we will need to use it in other places. Looks like this: 

Screenshot 2026-05-28 at 4 15 36 PM

This optional `children` prop gives us more stylistic control over the style select input

@badiuoanaalexandra  for review.


badiuoanaalexandra (approved):
LGTM

PR #257, opened 2026-08-07
Fix Toast right padding when closeButton is false with an action button

tristen:
When a `Toast` is rendered with `closeButton={false}` **and** an `action` button, an unwanted gap appears to the right of the action button. This fixes that


miafan23 (approved):
LGTM

